Issue #4, opened 2019-02-12
Cryptor module is not asynchronous


ll-aashwin-ll:
Following issue #18, the cryptor module ustilises node's 'crypto' module for aes ciphering. But the 'crypto' module only offers a sync api. 

Cryptor tries to expose an async method by using to process.nextTick() wrapped around sync calls, but this , as discussed in #18, does not actually make it async. It will still block the event loop. 

There is this library:
https://www.npmjs.com/package/@ronomon/crypto-async

But even it is not truly async. It uses a c++ add on, but on the main thread. So once again, it will block the event loop. It also does not support aes gcm which efs uses.

There is possible scope to develop a new truly async crypto (or at least aes) library utilising node's thread pool like the linked library mentions in 'some new ideas' section.

As for now, since there seems to be know suitable solution for an async crypto lib, efs will continue using the built-in sync crypto library.

CMCDragonkai:
I experienced this problem with regards to js-polykey. Basically any crypto is generally slow as it involves proper CPU computation, and crypto is not "IO". This is why crypto is generally not async in JS. In order to make our crypto high performance, we will need to integrate multi-threading/multi-processing into our system.

CMCDragonkai:
Another option is: https://github.com/PeculiarVentures/node-webcrypto-ossl

This aligns our crypto API usage with the WebCrypto API which is supported by browsers. Of course our system cannot work in browsers due to the lack of native filesystem.

While in NativeScript, none of this will work because they don't have a JS native crypto, instead they will have to call native Android crypto or native iOS crypto.

CMCDragonkai:
There are some cool libraries here: https://github.com/PeculiarVentures

CMCDragonkai:
This issue is relevant: https://github.com/PeculiarVentures/webcrypto/issues/1

CMCDragonkai:
I think we should also explore kbpgp's dependencies and the crypto libraries they use there.

CMCDragonkai:
Here are the dimensions we have to evaluate:

1. Sync vs Async - We need async in most cases, but a sync API would be nice to be easy to test
2. Concurrent vs Parallel - Async implies its either concurrent or parallel, we need libraries that can be truly parallel because even if it's concurrent, encryption is real computation and will slow down our platform if it's still single threaded
3. Platform support - Node, iOS, Android (browser support is unnecessary because native fs doesn't exist there anyway, browser support is nice for other possible uses)

[...]

PR #13, opened 2020-04-08
TypeScript migration

### LINKED ISSUE #14: Migrate to TypeScript and JavaScript-Demo Environment
The JavaScript demo environment has a typescript branch: https://github.com/MatrixAI/JavaScript-Demo/tree/typescript

We intend to migrate our projects to this, except for the ones that have reached stability like `js-virtualfs`.

But all the other polykey projects are going to be migrated over as well.

So there are 2 aims here:

1. Start using typescript + webpack
2. Use the same Nix structure as JS demo

Make sure that all tests have been migrated as well to Jest.

robert-cronin:
Fixes #14.


robert-cronin:
@CMCDragonkai

CMCDragonkai:
All PRs and MRs are meant to be linked to a particular issue that they are resolving. Currently there isn't an issue in this repo that is for the typescript migration.

But we should do this in stages.

Make this PR focus on migrating to typescript and only that (and to change to using the JavaScript-Demo architecture "typescript" branch).

Once that is done, we merge this, and we can put in the async issue into a separate MR.

CMCDragonkai:
I've added #14. I think you said the TS migration is already done. So can we merge that portion of work on Monday, and extract out the extra work on the async things into a new MR targeting #4.

package.json
```diff
@@ -1,58 +1,64 @@
...
-    "rollup-plugin-babel": "^4.0.2",
-    "rollup-plugin-commonjs": "^9.1.6",
-    "rollup-plugin-node-builtins": "^2.1.2",
-    "rollup-plugin-node-globals": "^1.2.1",
-    "rollup-plugin-node-resolve": "^3.3.0",
-    "virtualfs": "^2.2.0"
-  },
   "dependencies": {
-    "@babel/runtime": "^7.0.0"
+    "immutable": "^4.0.0-rc.12",
```
CMCDragonkai, line 32: Note that when using webpack... be ware of what are `dependencies` and what are `devDependencies`. Anything that webpack bundles can be put in `devDependencies`. Anything that `webpack` doesn't bundle is `dependencies`.

This is more important for applications than libraries, because for libraries, almost all external packages are `dependencies`, except for things that are explicitly typescript plugins or compile-phase tools.

CMCDragonkai:
You need to remember to set the metadata of PRs. See on the right.

CMCDragonkai:
I'll review this tonight. But after review, squash into 1 commit.

CMCDragonkai:
I've removed the `javascript-demo`. That is just an example from JS Demo. But also updated the `tsconfig.json` to match how we do it in other apps. Please try this.

Also can move the tests back into a `tests` directory? And if you can do any test based IO with `fs.mkdtemp` directory instead of inside the `test` directory.

CMCDragonkai:
I've moved `js-virtualfs` back into `dependencies` because it shouldn't be bundled in by default when we export commonjs modules.

Furthermore, because  the webpack config has changed as well. @robert-cronin you need to try rebuilding using webpack to see if the config works as normal as before.

CMCDragonkai:
Once this is merged, make sure to rebase #15 on this.

CMCDragonkai:
Confirm that builds still work, and check if `package.json` fields `main`, `module`, `browser` matches what is outputted in `dist` directory. And then if that's true. Squash and merge.

PR #15, opened 2020-04-20
Webworker async

### LINKED ISSUE #4: Cryptor module is not asynchronous

robert-cronin:
Fixes #4


CMCDragonkai:
This MR should be squashed and then rebased on master. Then you can proceed. Otherwise this won't be mergeable.

robert-cronin:
This PR is closed in favour of PR #19

robert-cronin:
This branch can be deleted in light of the recently merged #19

Issue #18, opened 2020-04-27
Node process context control


robert-cronin:
VFS has certain methods to control the file system context (e.g. chdir, setUid, setGid etc), but this is not exposed on the native nodejs fs module. It is instead controlled via the `process` module. In order to change the context on the lower fs, one would need access to the `process` module and this entails some extra research as to the best way to approach this. Essentially we want to be careful about tampering with the process context.

One solution to this is to ask for these methods as an additional parameter when the user passes in `upperDir` and `lowerDir`. The new constructor for EFS could look something like this:

```
const efs = new EncryptedFS(
  vfs,    // upperDir
  vfs,    // upperDir context control (for chdir, setUid, setGid, etc...)
  fs,     // lowerDir
  process // lowerDir context control (for chdir, setUid, setGid, etc...)
  ...
)
```

These context control objects have to conform to an interface to ensure that the relevant methods exist:

```
interface FSContextControl {
  chdir(...): void
  setUid(...): void
  setGid(...): void
}
```

But if we are going to use `process`, then we have to ensure that EFS is notified whenever the process context is changed externally to EFS (i.e. manually by user). This could be done by creating a proxy process that acts as an observer pattern and using this in EFS instead so that EFS is notified every time the cwd/gid/uid is changed in the process context. This should also go in usage notes and operator warnings to ensure correct usage by the end user.

robert-cronin:
I think this issue is largely completed as the latest release includes these context control methods.

PR #19, opened 2020-04-27
Webworkers for Async and Jest test migration

### LINKED ISSUE #4: Cryptor module is not asynchronous

### LINKED ISSUE #20: Documentation
Documentation needs to be created and should include:
- Overview (explanation of upper and lower directory/chunks etc)
- Installation
- Usage
- Warnings (e.g. for usage of context control actions like chdir and setuid etc.)

This can go into `README.md`

### LINKED ISSUE #21: Crypto - initiVector Variable Length
If an initVector is not provided to the `Crypto` class upon initialisation, a random vector is generated of length 16 bytes. Should this be variable length? Plausibly one could assume this functionality is already available to users who specify their own initiVector but there are methods that cannot be changed. My suggestion is to use the length of the existing initVector to drive any further creation of initVectors, this lets the user control the initVector length.

robert-cronin:
This PR is a combination of webworkers async and bringing the ava tests from VFS and migrating them to jest in EFS. This allows us to be somewhat certain that EFS exposes the same functionality as VFS.

Fixes #4 
Fixes #20 
Fixes #21


robert-cronin:
There are still a few tests that need investigating:
- streams tests
- tests that use setGid/setUid/chDir

CMCDragonkai:
Continue with the last few tests, I'll get back to you on the review next week.

CMCDragonkai:
We talked about the process context problem that real filesystems have. To model this in a VFS, we had to simulate these functions in the VFS constructor.

However since EFS is a hybrid of VFS and the real FS. Or any FS at all. Then it cannot expect that these operations (uid, gid, cwd) exist on the fs itself. They were just put into VFS for simulation purposes.

Therefore I suggest that you allow the caller to provide an object/interface to these functions. It's easy enough that VFS can be supplied twice and EFS can call this, and in the case of real fs, you supply the `process` object.

This means EFS takes 4 things:

1. The upper fs
2. The upper fs context
3. The lower fs
4. The lower fs context

So I'm guessing this in the TODO list. But you need to reference the issue for this problem in this PR so that way when we read it, it's possible to follow the hyperlinks.

CMCDragonkai:
Final thing is that I don't see any confirmation on whether the last tests you said are failing are still failing or we are just putting them into another issue to solve later.

package.json
```diff
@@ -24,18 +28,27 @@
   "jest": {
     "verbose": true
   },
+  "engines": {
+    "node": ">=8.7.0"
```
CMCDragonkai, line 32: Is this the Node version you are running?

package.json
```diff
@@ -24,18 +28,27 @@
...
+  },
   "dependencies": {
-    "virtualfs": "^2.2.0"
+    "lodash": "^4.5.0",
```
CMCDragonkai, line 35: I'd prefer that we didn't use lodash. These utility libraries are intended for applications, not library authors themselves.

src/Cryptor.ts
```diff
@@ -1,131 +1,160 @@
...
-import * as process from 'process';
+import * as fs from 'fs'
+import * as crypto from 'crypto'
+import * as process from 'process'
+import { spawn, Worker, ModuleThread } from 'threads'
+import { CryptorWorker } from './CryptorWorker'
 
 // TODO: function docs
 
+// interface CryptorParameters {
```
CMCDragonkai, line 9: These sorts of comments should be left out. If commented code is left in our master branch, there must be a good a reason for doing so. Consider this:

1. If they are not used anymore - delete them
2. If they are to be implemented in the future - turn it into an issue and delete them
3. If they are not understood - create an issue and delete them
4. If they are documentation - then it should be written in a doc format

src/Cryptor.ts
```diff
@@ -1,131 +1,160 @@
...
-  private _decipher: crypto.Decipher;
-	constructor(pass: string, initVector: Buffer = null, algo: string = 'id-aes256-GCM') {
-		this._algo = algo;
-    	this._initVector = initVector ? initVector : this.genRandomIVSync();
+	private _algorithm: string
+	private _initVector: Buffer
+	private _key: Buffer
+	private _cipher: crypto.Cipher
+	private _decipher: crypto.Decipher
+	private _useWebWorkers: boolean
```
CMCDragonkai, line 22: If you use `private` scoping. Do you need to use `_` as a prefix? Or is that convention in typescript? If it's not convention in TS, then private scope is enough without needing to also use `_` as a prefix.

src/Cryptor.ts
CMCDragonkai, line 9: These sorts of comments should be left out. If commented code is left in our master branch, there must be a good a reason for doing so. Consider this:

1. If they are not used anymore - delete them
2. If they are to be implemented in the future - turn it into an issue and delete them
3. If they are not understood - create an issue and delete them
4. If they are documentation - then it should be written in a doc format

package.json
CMCDragonkai, line 35: I'd prefer we didn't use lodash. These are meant for applications, not libraries.

src/Cryptor.ts
```diff
@@ -1,131 +1,160 @@
...
+	private _cryptorWorker?: ModuleThread<CryptorWorker>
+	constructor(
+		key: Buffer | string,
+		initVector: Buffer = crypto.randomBytes(16),
+		algorithm: string = 'aes-256-gcm',
+		useWebWorkers: boolean = false
+	) {
+		this._algorithm = algorithm
+		this._initVector = initVector
 		// TODO: generate salt ?
```
CMCDragonkai, line 32: For example here, if this is a real question, turn it into an issue.

robert-cronin:
The stream tests are integrated, there are still permissions tests left that we can put on a separate branch. 

The context control parameters are already added to the constructor but the way they are used in efs will have to be worked out. I propose this going in a separate issue as well, for now it make sense to merge first as the context control and permissions functionality have nothing or little to do with migration to jest.

Context control is already captured in issue #18. Probably worth adding a note about the extra note about the relevant tests.

src/Cryptor.ts
```diff
@@ -1,131 +1,160 @@
...
+		}
+		callback(null, buffer)
 	}
 
-	// TODO: needs iv param
-	decrypt(...args: Array<any>) {
-		let argSplit = this._separateCallback(args)
-		let cb = argSplit.cb;
-		let methodArgs = argSplit.args;
+	decryptSync(cipherBuf: NodeJS.ArrayBufferView, initVector?: Buffer): Buffer {
```
CMCDragonkai, line 69: Could you explain what does `NodeJS.ArrayBufferView` come from? I was under the assumption that native JS types which `ArrayBufferView` is is just global in TS. So you can just say `ArrayBufferView`.

And if you do use `NodeJS` what does this mean for browsers or non-nodejs environments?

src/CryptorWorker.ts
```diff
@@ -0,0 +1,28 @@
+import { expose } from "threads"
+import * as crypto from 'crypto'
+
+let _cipher: crypto.Cipher
```
CMCDragonkai, line 4: Why are these using the `let` declaration? Everything at module namepace should be `const`. We only use `let` in code we intend to reassign later.

src/CryptorWorker.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const _cryptorWorker = {
+  init(algorithm: string, key: Buffer, initVector: Buffer): void {
+		_cipher = crypto.createCipheriv(algorithm, key, initVector)
+		_decipher = crypto.createDecipheriv(algorithm, key, initVector)
+  },
+  updateCipher(algorithm: string, key: ArrayBuffer, initVector: ArrayBuffer, plainBuf: crypto.BinaryLike): Buffer {
+    return _cipher.update(plainBuf)
+  },
+  _resetCipher(algorithm: string, key: crypto.CipherKey, initVector: crypto.BinaryLike): void {
```
CMCDragonkai, line 15: What exactly is `_resetCipher` used for?

src/EncryptedFS.ts
```diff
@@ -15,127 +19,1340 @@ import VFS from 'virtualfs';
...
+}
+
+type PathLike = string | Buffer | URL
+type FileLike = PathLike | number
+type NoParamCallback = (err: NodeJS.ErrnoException | null) => void
+interface MakeDirectoryOptions {
+	recursive?: boolean,
+	mode?: number
+}
+export interface Stat {
```
CMCDragonkai, line 58: Both this interface `Stat` and `GeneralFilesystem` interface should be moved to their own files. I don't think these types should be defined here.

Furthermore I'm pretty sure TS has already an interface defined for Node's filesystem. Can't we use that?  That is, if you have the right type root. Can't you just use `a: fs`? See https://stackoverflow.com/questions/43048113/use-fs-in-typescript

This is just an experiment, but if we can do this we should be a lot leaner and not need to maintain this giant interface.

CMCDragonkai: One issue might be that some real fs functions we didn't implement. But we can deal with that by creating a stub function and then throwing an exception there. You can submit a PR to js-virtualfs to facilitate that.

src/constants.ts
```diff
@@ -0,0 +1,61 @@
+const constants = Object.freeze({
```
CMCDragonkai, line 1: I see that `virtualfs` is a dependency for js-encryptedfs.

Well if that's true, you can just use the constants defined inside there.

However I believe `virtualfs` **should not**  be a dependency of `js-encryptedfs` because EFS should just take any fs object in. Why does it need the virtualfs dependency at all?

tests/Cryptor.unit.test.ts
```diff
@@ -1,79 +1,96 @@
...
+import * as jest from 'ts-jest'
+import * as crypto from 'crypto'
 
 
 describe('Cryptor class', () => {
 	// Define initialization vector for all tests
-	const iv = crypto.randomBytes(16);
+	let initVector: Buffer
+	let key: Buffer
+	let cryptor: Cryptor
```
CMCDragonkai, line 10: Why did we call this `Cryptor`? That is such a strange name. Can we make it more obvious what this thing is. If it's our crypto utility, why not just name it `Crypto`?

todo
```diff
@@ -0,0 +1,11 @@
+1) Move EFS to using fdIndex and an internal file descriptor index
```
CMCDragonkai, line 1: These things should either:

1. Turned into issues
2. Put into the README.md (at the end after a hr line)

tsconfig.json
```diff
@@ -2,17 +2,19 @@
...
     "sourceMap": true,
-    // "strictNullChecks": true,
+    "strictNullChecks": true,
+    "allowJs": true,
     "noImplicitAny": false,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "module": "CommonJS",
-    "target": "ES2016",
+    "target": "ES2018",
```
CMCDragonkai, line 11: I thought we are standardizing around `ES2016`?

I guess it depends on our platform. How old can we go? ES2016 would be best for web apps. But since this is consumed by:

1. CLI
2. GUI
3. Mobile

Then perhaps we can go to later versions and get some performance boost?

tsconfig.json
```diff
@@ -2,17 +2,19 @@
   "compilerOptions": {
     "outDir": "./dist",
     "sourceMap": true,
-    // "strictNullChecks": true,
+    "strictNullChecks": true,
+    "allowJs": true,
```
CMCDragonkai, line 6: Why did we allow this?

webpack.config.js
```diff
@@ -2,30 +2,58 @@ const process = require('process');
...
+    filename: 'encryptedfs.js',
     library: 'encryptedfs',
     libraryTarget: 'umd',
   },
+  externals: {
+    virtualfs: "virtualfs",
+    threads: "threads",
+  },
+  devtool: "source-map",
+  devServer: {
```
CMCDragonkai, line 22: Remove the `devServer` this is not a web application.

webpack.config.js
```diff
@@ -2,30 +2,58 @@ const process = require('process');
...
   },
-  externals: [nodeExternals()],
+  externals: [
+    nodeExternals()
+  ],
+  plugins: [
+    new ThreadsPlugin({
+      globalObject: false
+    }),
+    new webpack.EnvironmentPlugin([
```
CMCDragonkai, line 52: Get rid of this too.

src/Cryptor.ts
robert-cronin: Microsoft recommends against using underscores: https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines
so I will remove them

package.json
robert-cronin: lodash usage was inconsequential and I've replaced it with native methods

src/Cryptor.ts
robert-cronin: This was just an initial attempt on my behalf to satisfy typescript in providing a type that feeds into the crypto library but we should be using the more relevant `Buffer` class

robert-cronin: this also applies to the lines where crypto.BinaryLike is specified and can be replaced with `Buffer | string`

src/CryptorWorker.ts
robert-cronin: as far as I know, this is just not possible with web workers, we have to pass algo/key/IV in order to recreate the same cryptor in the web workers context. I am unaware of any way to pass custom initialization parameters to threads.js so I've had to resort to a let statement and an `init` function.

src/EncryptedFS.ts
robert-cronin: If were going to move `Stat` and `GeneralFileSystem` then why not move all types and interfaces to a separate file like `util.ts`? I don't understand why just those two.

Just as in js-polykey, we could probably just do something like `lowerDir: typeof fs` but as you pointed out, the reason for the UpperDirectoryFileSystem interface is because VFS doesn't implement all of the same methods as the native fs module. I also think some methods might not conform to the file system api. I can open a PR on a new branch for this on vfs repo.

src/constants.ts
robert-cronin: yes `virtualfs` should go into `devDependencies` only for testing purposes. I thought the constants object was a nice little convenience that `virtualfs` provided for defining `mode` in `access` and `open` methods so I stole it from `virtualfs`.

tests/Cryptor.unit.test.ts
robert-cronin: well I can't comment on the original naming of `Cryptor`, but `Crypto` sounds more conventional. One issue might be a conflict with the `crypto` library when defining `let crypto: Crypto`.

src/EncryptedFS.ts
robert-cronin: I don't think we need to implement stub functions in `virtualfs` as I don't think typescript type checks for pure js objects, the error is thrown at run-time so as long as we don't use those unimplemented methods, its fine.

tests/Cryptor.unit.test.ts
robert-cronin: okay so there is no namespace error thrown as we don't import `crypto` in `EncryptedFS.ts`.

tsconfig.json
robert-cronin: this came up in our last meeting on polykey, it is recommended when using typescript. It requires you to explicitly specify an `undefined` or `null` type instead of existing in the domain of every type. This is advantageous when want to force a required parameter not to be null like `path` in `readdirSync`

webpack.config.js
robert-cronin: this was previously used to debug on the console in the browser, but it is replaced with the more appropriate `ts-node`

package.json
robert-cronin: node version was updated to `12.14.0`

src/Cryptor.ts
robert-cronin: Relevant issues have been extracted into either a new issue or README and extraneous comments have been deleted

src/CryptorWorker.ts
robert-cronin: it is used to reinitialise the cipher with a new `initVector`

tsconfig.json
robert-cronin: yes I think ES2018 is the best way to go.

todo
robert-cronin: I opted for option 2, they are in the README.

src/Cryptor.ts
robert-cronin: This has been turned into issue #22

src/CryptoWorker.ts
```diff
@@ -0,0 +1,28 @@
+import { expose } from "threads"
+import * as crypto from 'crypto'
+
+let cipher: crypto.Cipher
+let decipher: crypto.Decipher
+
+const cryptoWorker = {
+  init(algorithm: string, key: Buffer, initVector: Buffer): void {
+		cipher = crypto.createCipheriv(algorithm, key, initVector)
```
CMCDragonkai, line 9: There is literal tab characters. They should be spaces.

src/Streams.ts
```diff
@@ -0,0 +1,337 @@
+import { Buffer } from 'buffer'
+import { nextTick } from 'process'
+import { Readable, Writable } from 'stream'
```
CMCDragonkai, line 3: Change this to `import { Readable, Writable } from 'readable-stream';`. This is portable stream library.

src/Streams.ts
```diff
@@ -0,0 +1,337 @@
+import { Buffer } from 'buffer'
```
CMCDragonkai, line 1: Use https://github.com/feross/buffer instead of Node's buffer. And we need to ensure we are getting the buffer from the library.

package.json
```diff
@@ -24,18 +28,26 @@
...
   "dependencies": {
-    "virtualfs": "^2.2.0"
+    "threads": "1.4.0",
+    "virtualfs": "^2.2.0",
+    "errno": "^0.1.7"
   },
   "devDependencies": {
-    "@types/jest": "^24.9.1",
+    "@types/jest": "^25.0.0",
     "@types/node": "^13.13.1",
```
CMCDragonkai, line 41: The `"@types/node"` provides the fs type: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/fs.d.ts.

This should mean we do not need to create our own Filesystem interface.

Furthermore as long as we use the `fs` as a type, and not as an implementation. The typescript compilation should remove all references when it generates the JS and thus we won't have any portability problems.

src/index.ts
```diff
@@ -1,8 +1,7 @@
...
+
+export { Stat } from './util'
 
-// polyfills to be exported
-// $FlowFixMe: Buffer exists
 export { Buffer } from 'buffer';
-// $FlowFixMe: nextTick exists
-export { nextTick } from 'process';
\ No newline at end of file
+export { nextTick } from 'process';
```
CMCDragonkai, line 7: If you are not using `nextTick`, then this must be removed. And the `process` library as well.

CMCDragonkai: And since you are propagating the lowerfs async, you don't need to implement your own async wrapper, so we shouldn't need to have a `nextTick`.

CMCDragonkai: Propagating the lowerfs is `promisify`.

src/util.ts
```diff
@@ -0,0 +1,153 @@
...
+export type FileLike = PathLike | number
+export type NoParamCallback = (err: NodeJS.ErrnoException | null) => void
+export interface RmDirOptions {
+	recursive?: boolean
+}
+export interface MakeDirectoryOptions {
+	recursive?: boolean,
+	mode?: number
+}
+export interface Stat {
```
CMCDragonkai, line 25: We should get this from the `"@node/types"`. So this shouldn't exist just like what we talked about with respect to the `GeneralFilesystem`.

tsconfig.json
CMCDragonkai: Test not having this option and see if it works.

src/CryptoWorker.ts
CMCDragonkai: Check if your editor is respecting the `.editorconfig`. If you don't have one, get one from `JavaScript-Demo` repo.

tsconfig.json
robert-cronin: webpack, tsc and jest can all run without this option so I've removed it.

src/util.ts
robert-cronin: These redundant interfaces have been removed and replaced with types from nodes.

package.json
robert-cronin: this is now resolved by using the fs type on lower directory and an any type on the upper directory

src/Streams.ts
robert-cronin: Done, although I'm not sure how to ensure its using the buffer from the library. I have put this line into webpack config:
```
module.exports = {
  ...
  Buffer: require('buffer'),
  ...
}
```

src/Streams.ts
robert-cronin: Done. This can only be done with `const { Readable, Writable } = require('readable-stream')`

src/CryptoWorker.ts
```diff
@@ -0,0 +1,28 @@
+import { expose } from "threads"
+import * as crypto from 'crypto'
+
+let cipher: crypto.Cipher
```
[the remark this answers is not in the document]
robert-cronin: This has been done alongside a few extra things that pertain to the web workers code.

src/CryptoWorker.ts
```diff
@@ -0,0 +1,28 @@
...
+let decipher: crypto.Decipher
+
+const cryptoWorker = {
+  init(algorithm: string, key: Buffer, initVector: Buffer): void {
+		cipher = crypto.createCipheriv(algorithm, key, initVector)
+		decipher = crypto.createDecipheriv(algorithm, key, initVector)
+  },
+  updateCipher(algorithm: string, key: ArrayBuffer, initVector: ArrayBuffer, plainBuf: crypto.BinaryLike): Buffer {
```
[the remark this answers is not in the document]
robert-cronin: Fixed alongside making the functions have lexical scope

src/CryptoWorker.ts
robert-cronin: All files reformatted according to .editorconfig

CMCDragonkai:
You couldn't import it? Perhaps you're missing some tsconfig or webpack config?

On 8 May 2020 13:55:39 GMT+10:00, Robbie Cronin <notifications@github.com> wrote:
>@robert-cronin commented on this pull request.
>
>
>
>> @@ -0,0 +1,337 @@
>+import { Buffer } from 'buffer'
>+import { nextTick } from 'process'
>+import { Readable, Writable } from 'stream'
>
>Done. This can only be done with `const { Readable, Writable } =
>require('readable-stream')`
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/MatrixAI/js-encryptedfs/pull/19#discussion_r421922730

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

CMCDragonkai:
can't we use fs type on both?

On 8 May 2020 12:59:42 GMT+10:00, Robbie Cronin <notifications@github.com> wrote:
>@robert-cronin commented on this pull request.
>
>
>
>>    },
>   "devDependencies": {
>-    "@types/jest": "^24.9.1",
>+    "@types/jest": "^25.0.0",
>     "@types/node": "^13.13.1",
>
>this is now resolved by using the fs type on lower directory and an any
>type on the upper directory 
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/MatrixAI/js-encryptedfs/pull/19#discussion_r421909002

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/Streams.ts
robert-cronin: this is now working

src/index.ts
robert-cronin: callbacks have been promisified and we are now using `async` implicit promises in lieu of `nextTick`

robert-cronin:
> You couldn't import it? Perhaps you're missing some tsconfig or webpack config?
> […](#)
> On 8 May 2020 13:55:39 GMT+10:00, Robbie Cronin ***@***.***> wrote: @robert-cronin commented on this pull request. > @@ -0,0 +1,337 @@ +import { Buffer } from 'buffer' +import { nextTick } from 'process' +import { Readable, Writable } from 'stream' Done. This can only be done with `const { Readable, Writable } = require('readable-stream')` -- You are receiving this because your review was requested. Reply to this email directly or view it on GitHub: [#19 (comment)](https://github.com/MatrixAI/js-encryptedfs/pull/19#discussion_r421922730)
> -- Sent from my Android device with K-9 Mail. Please excuse my brevity.

I was able to import it with `import { Buffer } from 'buffer/'`

CMCDragonkai:
Can you squash this MR down to 1 or a few commits? It's time to merge this.

robert-cronin:
This is now squashed into 3 distinct commits: webworkers, jest migration and docs. Ready for merging :+1:

CMCDragonkai:
Congrats.

CMCDragonkai:
This project isn't finished yet though! We still need to polish and ensure our crypto works properly.

robert-cronin:
yes, I am researching GCM now, it will be essential to ensure EFS encryption/decryption conforms to GCM standards

PR #23, opened 2020-05-13
Prepare for npm publish and also reworking the cryptoworker and additional tests

### LINKED ISSUE #16: Integrate Generic Crypto library
ATM this is using Node crypto. This is not portable to the other systems that we intend to use js-encryptedfs with.

Now this isn't using PGP, so we don't use kbpgp or openpgp.js. However we are using AES.

It seems we have a couple of choices here:

* https://github.com/keybase/node-nacl
* https://github.com/brix/crypto-js
* https://github.com/PeculiarVentures/webcrypto

I'm not sure if the NaCL library fits our requirements.

Furthermore nativescript will require a completely different implementation too. So we would likely need to tackle that later.

robert-cronin:
This is to prepare and eventually do npm publish
Fixes #16


tests/Cryptor.unit.test.ts
```diff
@@ -1,124 +1,139 @@
 import Crypto from '../src/Crypto'
-import { Buffer } from 'buffer/'
+import * as nodeCryptoLib from 'crypto'
```
CMCDragonkai, line 2: Why not just `import * as crypto from 'crypto'`?

CMCDragonkai:
I think to go through the crypto code we would need to have a meeting. It would be difficult for me to review it on my own.

tests/Cryptor.unit.test.ts
robert-cronin: this was to distinguish from our own `let crypto: Crypto` statements, we could rename either and I don't think it would matter

robert-cronin:
> I think to go through the crypto code we would need to have a meeting. It would be difficult for me to review it on my own.

I agree, there is quite a bit to go through here, let me know when is best :+1:

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
...
-  }
+    // Random salt
+    const salt = this.cryptoLib.randomBytes(cryptoConstants.SALT_LEN)
 
-  decryptSync(cipherBuf: Buffer, initVector?: Buffer): Buffer {
-    if (initVector && (initVector !== this.initVector)) {
-      this.resetDecipherSync(initVector!)
-    }
+    // Reset cipher
+    this.resetSync(this.masterKey, initVector, salt)
```
CMCDragonkai, line 58: Main thing here is to add a comment explaining why it's important to "reinitialize" with a new key and `initVector`. It's important to cover all the edgecases of random access read/write and document the behaviour.

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
...
-  }
-
-  private async resetCipher(initVector: Buffer) {
-    this.initVector = initVector
-    if (this.useWebWorkers && this.cryptorWorker) {
-      return await this.cryptorWorker.init(this.algorithm, this.key, initVector)
-    } else {
-      this.cipher = crypto.createCipheriv(this.algorithm, this.key, this.initVector)
+      // Construct chunk
+      return Buffer.concat([salt, initVector, authTag, encrypted])
```
CMCDragonkai, line 93: A "packet" diagram for this would be useful.

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
...
+
+  private async delay(ms: number) {
+    return new Promise( resolve => setTimeout(resolve, ms) )
+  }
+  private async waitForCryptoWorkerInit(): Promise<boolean> {
+    for (let trial=0; trial < 10; trial++) {
+      if (this.cryptoWorker) {
+        return true
+      } else {
+        await this.delay(100)
```
CMCDragonkai, line 147: EFS doesn't concern itself with the OS problems. And not being able to start a cryptoworker sounds an OS problem. And so here we should expect that the worker does start. We can have a configurable timeout (that's a parameter that can be set by the user when they construct EFS). And if this timeout passes by, then we can throw an exception. The timeout can be set by default to `null` which would mean that we just wait forever.

Why doesn't threads.js give us a "promise" based interface to starting a worker? `startWorker.then(() => ...)`.

One way to deal with this is to make the constructor block until it has constructed the worker. And this seems to make sense, because we wouldn't start encrypting anything until we have a worker ready. We can use a white loop to do this. But we should try to have a "sleep".

One issue with this, is that in Polykey we expect to have many vaults and each vault is backed by EFS. In this current implementation, we would have a webworker per vault. Doing this only makes sense when the process has some process stateful context that matters to that process. Our webworker process is just doing pure computation. And we pass everything it needs when it does it. Therefore it seems like we should have the ability to share a webworker pool across multiple instances of EFS.

To do this, we can abstract the "initialisation" of workers to a separate utils file. And let js-polykey call that ahead of time to create the workers (and then in this case, it can actually do it asynchronously). And then chain that with the code that constructs the EFS. This means one can pass `worker_pool` (and can consist of just 1 worker) to the EFS constructor and make it use it. If no worker pool is passed, we don't use workers at all.

src/CryptoWorker.ts
```diff
@@ -1,27 +1,38 @@
...
+  const tag = cipher.getAuthTag()
 
-function decryptBuf(encryptedBuf: Buffer): Buffer {
-  return Buffer.from(decipher.update(encryptedBuf))
+  // Construct chunk
+  return Buffer.concat([salt, initVector, tag, encrypted])
 }
 
-export {init, encryptBuf, decryptBuf}
+function decryptChunk(chunkBuffer: Buffer, masterKey: Buffer, algorithm: 'aes-128-gcm' | 'aes-192-gcm' | 'aes-256-gcm'): Buffer {
```
CMCDragonkai, line 20: Since the underlying impl is shared between `Crypto.ts` and `CryptoWorker.ts`, it's possible to abstract this into a shared `CryptoUtils`. Not necessary since its so small.

tests/Cryptor.unit.test.ts
CMCDragonkai: So we can leave `crypto` to mean Node's crypto. As that's recognizable easily. But then our specific form of crypto can be prefixed as `EncryptedFSCrypto`.

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
-import * as crypto from 'crypto'
+import { CryptoInterface, Cipher, Decipher, AlgorithmGCM, cryptoConstants, deconstructChunk } from './util'
 import { spawn, Worker, ModuleThread } from 'threads'
-import { Buffer } from 'buffer/'
 
 export default class Crypto {
```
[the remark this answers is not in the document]
robert-cronin: Everything is now exported at file bottom except `CryptoInterface`. I am not sure why but typescript doesn't allow you to export interfaces anywhere other than where they are defined

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
-import * as crypto from 'crypto'
+import { CryptoInterface, Cipher, Decipher, AlgorithmGCM, cryptoConstants, deconstructChunk } from './util'
```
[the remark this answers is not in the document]
robert-cronin: Done

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
...
 import { spawn, Worker, ModuleThread } from 'threads'
-import { Buffer } from 'buffer/'
 
 export default class Crypto {
-  private algorithm: string
-  private initVector: Buffer
-  private key: Buffer
-  private cipher: crypto.Cipher
-  private decipher: crypto.Decipher
+  private masterKey: Buffer
```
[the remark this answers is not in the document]
robert-cronin: I have added some ideas along these lines to the relevant issue: https://github.com/MatrixAI/js-polykey/issues/22

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
...
+  private algorithm: AlgorithmGCM
+  // Web workers
   private useWebWorkers: boolean
-  private cryptorWorker?: ModuleThread
+  private cryptoWorker?: ModuleThread
+  // Sync members
+  private cipher: Cipher
+  private decipher: Decipher
```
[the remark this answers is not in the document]
robert-cronin: Done

src/Crypto.ts
robert-cronin: Since the cipher and decipher have been moved to the encrypt and decrypt methods, this is no longer ambiguous. I've added documentation to the functions and a note in to make it a bit more obvious.

src/Crypto.ts
robert-cronin: I've added a chunk diagram to the beginning of the crypto class

src/CryptoWorker.ts
robert-cronin: I have just added it to the `Crypto.ts` and exported it for use in our web worker file

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
...
       })
     }
   }
 
-
-  encryptSync(plainBuf: string | Buffer, initVector?: Buffer): Buffer {
-    if (initVector && (initVector !== this.initVector)) {
-      this.resetCipherSync(initVector!)
+  ///////////
+  // Reset //
```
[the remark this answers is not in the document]
robert-cronin: Done.

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
...
-import { Buffer } from 'buffer/'
 
 export default class Crypto {
-  private algorithm: string
-  private initVector: Buffer
-  private key: Buffer
-  private cipher: crypto.Cipher
-  private decipher: crypto.Decipher
+  private masterKey: Buffer
+  private algorithm: AlgorithmGCM
```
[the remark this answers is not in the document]
robert-cronin: Done, this is hardcoded for now

src/Crypto.ts
robert-cronin: So it turns out the worker pool initialisation is synchronous! so none of this delay stuff. So I think a util function exposed to the user that initialises the pool first is redundant and we can just leave it in there. But I did it anyway because as you said, multiple EFS instances will want to share the same worker pool.

src/Crypto.ts
```diff
@@ -1,150 +1,152 @@
...
+  // Web workers
   private useWebWorkers: boolean
-  private cryptorWorker?: ModuleThread
+  private cryptoWorker?: ModuleThread
+  // Sync members
+  private cipher: Cipher
+  private decipher: Decipher
+  // Crypto lib
+  private cryptoLib: CryptoInterface
```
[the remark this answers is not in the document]
robert-cronin: We can make it general later (for the web workers) down the line by actually providing an interface and a byo crypto worker but yeah for now this can stay like this

tests/Cryptor.unit.test.ts
robert-cronin: Done

robert-cronin:
I've added quite a few bisimulation tests that test all the way from 1-3 blocks, 1-2 rounds of read/write as well as bisimulation tests for `readFileSync` and `writeFileSync`.

src/Crypto.ts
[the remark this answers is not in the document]
CMCDragonkai: That doesn't seem right. Are you sure about this? I've defined interfaces many times and had them exported.

CMCDragonkai:
Ok I'm merging this.

PR #25, opened 2020-05-20
Add linting with ESLint and Prettier

### LINKED ISSUE #24: Integrate prettier and ESlint for CI
Use https://gitlab.com/MatrixAI/js-encryptedfs Gitlab will perform CI for us and we can then reformat according to prettier and ESLint.

robert-cronin:
Fixes #24

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,20 @@
+variables:
+  GIT_SUBMODULE_STRATEGY: recursive
+
+cache:
+  key: ${CI_COMMIT_REF_SLUG}
+  paths:
+    - node_modules/
+
+stages:
+  - check
+
+before_script:
+  - node --version
+  - npm install
+
+prettier:
+  stage: check
+  image: node:12
+  script:
+    - npm run lint
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "semi": true,
+  "trailingComma": "all",
+  "singleQuote": true,
+  "printWidth": 120,
+  "tabWidth": 2
+}
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "develop": "webpack --mode=development --watch & webpack-dev-server --mode=development & wait",
     "build": "webpack --mode='production'",
     "prepublish": "tsc",
+    "lint": "eslint '{src,tests}/**/*.{js,ts}' --quiet --fix",
     "test": "jest",
     "dos2unix": "dos2unix dist/*",
     "chmod": "chmod u+x dist/bin/*"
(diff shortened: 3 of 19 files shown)


CMCDragonkai:
Can you make prettier or eslint set to 2 space indent?

.editorconfig
```diff
@@ -6,3 +6,8 @@ indent_style = space
 indent_size = 2
 insert_final_newline = true
 trim_trailing_whitespace = true
+editor.codeActionsOnSave = {
```
CMCDragonkai, line 9: What does this do? Are you sure we need this? I usually prefer that the editor doesn't do this automatically be instead through an explicit CLI command and git add, git commit.

.gitlab-ci.yml
```diff
@@ -0,0 +1,22 @@
...
+  key: ${CI_COMMIT_REF_SLUG}
+  paths:
+    - node_modules/
+
+stages:
+  - check
+
+before_script:
+  - node --version
+  - npm install --global prettier eslint-config-prettier eslint-plugin-prettier
```
CMCDragonkai, line 14: I realized since your prettier job is already installing prettier. Then there's no need to do this ahead of time. This is only used if there are deps being used by ALL jobs. Not all jobs require prettier right?

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "semi": true,
+  "trailingComma": "all",
+  "singleQuote": true,
+  "printWidth": 120,
+  "tabWidth": 4
```
CMCDragonkai, line 6: 2 space indent!

package.json
```diff
@@ -22,6 +22,7 @@
     "develop": "webpack --mode=development --watch & webpack-dev-server --mode=development & wait",
     "build": "webpack --mode='production'",
     "prepublish": "tsc",
+    "lint": "eslint '{src,tests}/**/*.{js,ts}' --quiet --fix",
```
CMCDragonkai, line 25: Is eslint calling prettier?

CMCDragonkai:
BTW gitlab is reporting a failed pipeline.

.editorconfig
robert-cronin: Yes I agree, this should be an explicit action we trigger on the CLI

.prettierrc
robert-cronin: No blind copy and paste! gotcha, this is fixed

package.json
robert-cronin: Yes prettier can be set up inside `.eslintrc` using the folloing extends command:
```
  "extends": [
    ...
    "prettier/@typescript-eslint",
    "plugin:prettier/recommended"
  ],
```

.gitlab-ci.yml
robert-cronin: Yeah that makes perfect sense, I've put it in the relevant job

.gitlab-ci.yml
```diff
@@ -0,0 +1,20 @@
...
+  key: ${CI_COMMIT_REF_SLUG}
+  paths:
+    - node_modules/
+
+stages:
+  - check
+
+before_script:
+  - node --version
+  - npm install
```
CMCDragonkai, line 14: We only need dev dependencies. But you can do that later, as I'm going to merge this.

PR #33, opened 2020-06-30
Conform to fs interface

### LINKED ISSUE #30: readdirSync doesn't return anything that isn't loaded in memory
I think for now its sufficient to just have readdir propagate the lower directory method. It is working this way for the async `readdir` but the sync method `readdirSync` has been mistakenly left as propagating the upper directory method.

### LINKED ISSUE #32: Adapt EFS to the File System Interface
We should adapt EFS to the [File System Interface](https://nodejs.org/api/fs.html).

This is pertinent to making EFS work with isomorphic git in [js-polykey](https://github.com/MatrixAI/js-polykey). All of our current methods work by promises but are exposed in the main object whereas the fs interface describes them as being part of a `promises` API exposed at `fs`.`promises`.

Another thing is `lstat` does not work properly yet as it only propagates the `upperDir` method and it might not be loaded. This should be changed to propagating the `lowerDir` method.

related to https://github.com/MatrixAI/js-polykey/pull/43#discussion_r446655767

### LINKED ISSUE #34: lstat/lstatSync should propagate lowerDir
For now, the lstat functions should just propagate the lowerDir method. Right now they propagate the upperDir and that is not working well with isomorphic-git on js-polykey.

Until we can get the upper and lower directories agreeing on file stats, it should stay this way.

robert-cronin:
This issue will make EFS conform to the fs interface by, at a minimum, separating promise based methods into a separate `promise` API. This is motivated by the fact that we need a callback wrapper over EFS in order to work with isomorphic-git in `js-polykey`. 

Fixes #30 
Fixes #32 
Fixes #34


CMCDragonkai:
A quick look, but I'm going to assume you did this well. So merge when you are ready.

robert-cronin:
there is still a few minor things I need to do like update the testing but I will merge after that

robert-cronin:
This has now been updated to callback and promise methods according to the nodejs fs interface. It works well in isomorphic git as a `PromiseClient` as well. I will merge this shortly.

Issue #50, opened 2021-09-20
Benchmark & Load Test EncryptedFS


CMCDragonkai:
Right now we only have benchmarks for crypto operations and the DB operations.

To properly assess what is the optimal block size to use (probably more than 4 KiB, but definitely less than 1 MiB), we have to do a proper load test on the EFS instance.

This requires performing a sequential set of operations vs parallel set of operations.

With the tests done in https://github.com/MatrixAI/js-polykey/issues/209, we can see that due to the speed for encryption/decryption, using a worker thread will always be slower cause calling workers have 1.5ms overhead.

The speed up can only happen when the there are multiple worker threads being used, so when there are multiple parallel operations occurring either through batching and/or multiple EFS reads/writes.

Benny is not the right tool to do load testing, but we may be able to adapt it for proper load testing.

Otherwise we can use a better load testing tool like artillery, but it needs to be easy to integrate, and not go through the network. This is not an HTTP server. It's a library.

Remember as long as the main thread is doing other things, and that the work is more expensive than call overhead divided by the number of workers, then we should have a profitable use of worker threads.

### Additional Context

* https://github.com/MatrixAI/js-polykey/issues/209

### Tasks

1. [x] - Investigate how to use benny to create
2. [x] - Investigate other load testing tools like benny
3. [x] - Write a load test involving EFS, random read/write operations vs sequential read/write operations
4. [ ] - Consider how these benchmarks interact with PK usage (with vault committing protocol)
5. [ ] - Add additional test for using `WorkerManager` inside `EncryptedFS`

CMCDragonkai:
Preliminary load testing has already been done for EFS. Of course we can see that workers are quite a bit slow, and for single core usage, it's never worth it to use workers. It's only when there are many other concurrent operations where it becomes worth it. The exact calculation is located in another issue I forgot about.

For now I'm going to close this issue until we have a more pressing need to benchmark.

PR #52, opened 2021-09-23
Loading and storing raw data buffers in the database

scottmmorris:
### Description

This PR will change the data buffers to be loaded and stored into the db using the raw parameter to avoid the unnecessary cost of JSON serialisation and deserialisation

### Tasks

1. [x] Introduce the raw parameter into the data transaction calls to the db
2. [x] Run the new benchmark tests with the update


src/inodes/INodeManager.ts
```diff
@@ -743,14 +743,14 @@ class INodeManager {
     for await (const data of dataDb.createReadStream(options)) {
       // This is to account for the case where a some blocks are missing in a database
       // i.e. blocks 0 -> 3 have data and a write operation was performed on blocks 7 -> 8
-      while (inodesUtils.unbufferId((data as any).key) !== blockCount) {
+      while (inodesUtils.unbufferId((data as any).key) != blockCount) {
```
CMCDragonkai, line 746: Why did you change this? The `!==` should be used for anything that is not a null comparison.

src/inodes/INodeManager.ts
```diff
@@ -743,14 +743,14 @@ class INodeManager {
...
         yield Buffer.alloc(blockSize);
       }
-      const plainTextData = await this.db.deserializeDecrypt<string>(
+      const plainTextData = await this.db.deserializeDecrypt(
```
CMCDragonkai, line 749: Does `raw` being `true` meant that we don't use the generic type parameter? I think it's because we're always expecting a `Buffer` right?

CMCDragonkai:
All tests passed and lintfixed?

src/inodes/INodeManager.ts
scottmmorris: I copied some of the code over so that must have been in the old code, Ill fix that now

src/inodes/INodeManager.ts
scottmmorris: Yeah if it is raw then it will always return a buffer

src/inodes/INodeManager.ts
```diff
@@ -769,15 +769,15 @@ class INodeManager {
...
     for await (const data of dataDb.createReadStream(options)) {
       key = inodesUtils.unbufferId((data as any).key);
-      value = await this.db.deserializeDecrypt<string>(
+      value = await this.db.deserializeDecrypt(
         (data as any).value as Buffer,
-        false,
+        true,
       );
     }
-    if (value == null || key == null) {
```
scottmmorris, line 777: Im fixing this up too, changing it back to null

scottmmorris:
Yep, linted and all tests passing with the latest commit

CMCDragonkai:
Ok please squash as well. I'll add my bench marks on top.

scottmmorris:
Have squashed

CMCDragonkai:
New benchmarks!

image

image

image

CMCDragonkai:
Compared to the master:

image

image

image

CMCDragonkai:
Looks like it's about up to 10% improvement in get and put speeds.

CMCDragonkai:
Sorry I just realised that the benchmarks shouldn't change much because we don't actually have EFS benchmarks yet. That's still pending due to #50. So ignore the above screencaps.

PR #54, opened 2021-09-27
Implementation of chroot function

scottmmorris:
### Description
This PR will introduce the `chroot` function into EFS. This function will return a reference to the existing EFS object with the root iNode changed to a specified path.

### Tasks

1. [x] Implement `chroot` function
2. [x] Handle any error cases
3. [x] Write tests for the new functionality
4. [x] Fix up any bugs encountered in testing

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/EncryptedFS.ts
+++ b/src/EncryptedFS.ts
@@ -243,6 +243,59 @@ class EncryptedFS {
     this.db.unsetWorkerManager();
   }
 
+  public async chroot(path: string): Promise<EncryptedFS>;
+  public async chroot(
+    path: string,
+    callback: Callback<[EncryptedFS]>,
+  ): Promise<void>;
+  public async chroot(
+    path: string,
+    callback?: Callback<[EncryptedFS]>,
+  ): Promise<EncryptedFS | void> {
+    return utils.maybeCallback(async () => {
+      if (!this._running) {
+        throw new errors.ErrorEncryptedFSNotRunning();
+      }
+      path = this.getPath(path);
+      const navigated = await this.navigate(path, true);
+      const target = navigated.target;
+      if (!target) {
+        throw new errors.ErrorEncryptedFSError({
+          errno: errno.ENOENT,
+          path: path as string,
+        });
+      }
+      await this.iNodeMgr.transact(
+        async (tran) => {
+          const targetType = (await this.iNodeMgr.get(tran, target))?.type;
+          const targetStat = await this.iNodeMgr.statGet(tran, target);
+          if (!(targetType === 'Directory')) {
+            throw new errors.ErrorEncryptedFSError({
+              errno: errno.ENOTDIR,
+              path: path as string,
+            });
+          }
+          if (!this.checkPermissions(constants.X_OK, targetStat)) {
+            throw new errors.ErrorEncryptedFSError({
+              errno: errno.EACCES,
+              path: path as string,
+            });
+          }
+        },
+        target ? [target] : [],
+      );
+      return new EncryptedFS({
+        db: this.db,
+        iNodeMgr: this.iNodeMgr,
+        fdMgr: this.fdMgr,
+        rootIno: target,
+        blockSize: this.blockSize,
+        umask: this.umask,
+        logger: this.logger,
+      });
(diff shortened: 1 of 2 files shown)


CMCDragonkai:
Tests need to be written for symlinks and also paths that use `/../` anywhere to prevent the user escape the chroot.

CMCDragonkai:
This should affect the root inode and current directory state. You might need to generalise the protected constructor of `EncryptedFS`.

scottmmorris:
After some testing, I will need to change more than just the properties of the encrypted file system. I needed to change these lines in the `navigate` function.

```
        if (parse.segment === '..' && curdir === this.rootIno) {
          target = curdir;
        } else {
          target = await this.iNodeMgr.dirGetEntry(tran, curdir, parse.segment);
        }
```

scottmmorris:
Are hardlinks able to escape the chroot @CMCDragonkai ? The problem is that with hardlinks is that it doesn't use the navigate function

CMCDragonkai:
Yes hardlinks are fine. I don't think it's a problem.

But if you're in the chroot it won't be possible to hardlink something 
outside the chroot.

Only if you are in the root context would you be able to hardlink 
between vaults. And there isn't a secret operation to do this.

scottmmorris:
Ready for merge

CMCDragonkai:
Do you think chroot needs to be async? It seems like something that can be done synchronously.

CMCDragonkai:
Oh I see you're using a transaction nvm.

src/EncryptedFS.ts
```diff
@@ -3228,7 +3281,11 @@ class EncryptedFS {
     let nextDir, nextPath, target, targetType;
     await this.iNodeMgr.transact(
       async (tran) => {
-        target = await this.iNodeMgr.dirGetEntry(tran, curdir, parse.segment);
+        if (parse.segment === '..' && curdir === this.rootIno) {
```
CMCDragonkai, line 3284: Can you explain how this works?

CMCDragonkai: This seems to rely on the idea that the `curdir` would be always set to `this.rootIno`.

But the `curdir` could be changed to other locations. That's the point of the `curdir`.

Can you add a test for what happens when the `curdir` is changes to other locations?

scottmmorris: If the current directory inode index is equal to the root iNode that is set at construction and the current parsed segment is going up a directory (..) then it will set the next target as the current directory (not moving up a directory)

scottmmorris: otherwise it will go up a directory as normal. This hasn't affected any of the other tests

CMCDragonkai: But the `curdir` index can be changed. It's not something that is static.

scottmmorris: the navigate function will iterate through dirs

CMCDragonkai: Oh I must have been confused. The `curdir` here is not the `CurrentDirectory`, it's local to this navigation recursion?

scottmmorris: yep correct

CMCDragonkai: At any case you should have a test to see what happens if the `CurrentDirectory` were to be changed. So prevent users from changing current directory above the chroot.

Also what happens if the current directory is inside another chroot.

One more test for nesting chroots. What if you chroot, and then chroot again.

CMCDragonkai: When all those pass, then you should be able to merge.

Do you have the npm permissions to publish?

CMCDragonkai: Remember:

```
npm patch minor
npm run build
git push
git push --tags
npm publish --access public
```

CMCDragonkai: Updated the above commands.

scottmmorris: Yeah I should have the right permissions

PR #56, opened 2021-10-08
Reloading EFS (iNodeMgr) state and introducing recursive directory options

scottmmorris:
### Description

This PR will fix two issues:
- Reloading EFS State: Currently, if an EFS object is created and then a second EFS instance with the same arguments is created an error will be thrown about duplicate root iNodes. Instead, this should set the root iNode as the existing root iNode and proceed with no errors.
- Recursive directory options: In removing a directory, there is currently no way to recursively delete a directory which will be fixed here. Also, there are currently two functions to create directories `mkdir` and `mkdirp` (second is for recursively creating dirs) while there should really only be one functions with the option to recursively create a directory 

### Tasks

1. [x] Merge `mkdir` and `mkdirp` functionality into one function 
2. [x] Add functionality to recursively delete a directory in `rmdir`
3. [x] Allow EFS to be loaded with existing state
4. [x] Expand testing for recursively deleting a directory
5. [x] Expand basic tests for EFS state

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


scottmmorris:
Final thing to do here is to add some testing in for recursively deleting directories

CMCDragonkai:
Does that mean `mkdirp` is gone?

Btw remember to squash when the CI passes too.

scottmmorris:
Yep `mkdirp` is gone I have merged the functionality into one `mkdir` function with a recursive option

CMCDragonkai:
Regenerate the docs please.

Then once merged on master.

```
npm version minor
npm run build
git push
git push --tags
npm publish --access public
```

CMCDragonkai:
Nice.

PR #57, opened 2021-11-30
Implementing a db key validity check on starting the encrypted fs

scottmmorris:
### Description

There needs to be a more explicit way of determining if the db key is valid for the encrypted fs. At the moment, the efs start up continues until it tries to decrypt existing state and cannot because the key is not valid.

To overcome this, a similar pattern to sentinel species will be used. On creation of the efs, we encrypt a known string into the db at the root level. Then on subsequent constructions, we access the string and if we cannot decrypt it then we know that the vault key is invalid and we therefore throw an error. If the string does not exist in the db then we start and put the known string in because we know its a new db.

### Tasks

1. [x] Implement the key validity checks
2. [x] Add new testing for the key validity checks

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/src/EncryptedFS.ts
+++ b/src/EncryptedFS.ts
@@ -14,7 +14,7 @@ import type { OptionsStream } from './streams';
 
 import { code as errno } from 'errno';
 import Logger from '@matrixai/logger';
-import { DB } from '@matrixai/db';
+import { DB, errors as dbErrors } from '@matrixai/db';
 import CurrentDirectory from './CurrentDirectory';
 import Stat from './Stat';
 import { INodeManager, errors as inodesErrors } from './inodes';
@@ -90,6 +90,7 @@ class EncryptedFS {
         logger: logger.getChild(DB.name),
       });
     }
+    await EncryptedFS.setupCanary(db);
     iNodeMgr =
       iNodeMgr ??
       (await INodeManager.createINodeManager({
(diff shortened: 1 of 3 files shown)


src/EncryptedFS.ts
```diff
@@ -89,6 +89,34 @@ class EncryptedFS {
         },
         logger: logger.getChild(DB.name),
       });
+      // Perform db key validation
+      const keyValidationDB = await db.level('keyValidation');
```
CMCDragonkai, line 93: No need for an additional level. Just use `db` the root level is empty.

src/EncryptedFS.ts
```diff
@@ -89,6 +89,34 @@ class EncryptedFS {
...
+      try {
+        const canary = await keyValidationDB.get('canary-key');
```
CMCDragonkai, line 95: I suggest just `canary` as the key is sufficient.

src/EncryptedFS.ts
```diff
@@ -89,6 +89,34 @@ class EncryptedFS {
...
+        // Try to decrypt the stored value
+        if (!(await utils.decrypt(dbKey!, utils.toArrayBuffer(canary)))) {
+          throw new errors.ErrorEncryptedFSKey();
+        }
+      } catch (e) {
```
CMCDragonkai, line 100: Separate out this logic:

```ts
// uses the root level that's already available
try {
  const deadbeefData: Buffer = await db.db.get('canary');

  try {
    const deadbeef = await db.deserializeDecrypt(deadbeefData, false);
    deadbeef === 'deadbeef';
  } catch(e) {
    // see what happens here, try different things
  }

} catch (e) {
  if (e.notFound) {
    // it doesn't exist, so you just put it and proceed
    await db.put([], 'canary', 'deadbeef');
  }
  throw e;
}
```

This initial canary check can be a separate protected function. You can call it `setupCanary`.

scottmmorris: I made it a static function because this is being done inside the static create function

scottmmorris: fixed now

CMCDragonkai:
This is cool!

But we realised we should probably do this against the `js-db` instead so that everything benefits from this. In `js-db` it would have to create a "_meta" level just for this stuff though and safeguard so that users don't try to create a level like so.

Issue #53, opened 2021-09-26
Identify and Eliminate Unscalable Operations with Lazy GC


CMCDragonkai:
### Specification

Some operations are "unscalable". For example to delete a file, one has to iterate over the entire file blocks and batch up all the delete operations and then execute them. This is due to the structure of how we are storing each file block as a separate key value entry in a sublevel.

If the file is large, this can take a long time. Furthermore deletion of files should be a lot faster than this. Real filesystems do not need to read the entire file just to delete it.

In order to solve this problem one way is to use lazy garbage collection. When deleting a file, it is marked as no longer accessible. The is already possible since file accessibility is determined by hardlinks which are persisted in the directory inodes.

However what is going to actually delete the key value entries? Real filesystems can simply mark filespace as reusable. Unfortunately this is not possible with a key value structure, and it would just get quite complex.

Therefore, we may instead of a garbage collection process that runs as a stream/loop and quietly deletes un-used entries in the background continuously as the EFS is being used. Right now we have something similar to this in order to remove zombie inodes in our gc sublevel.

Recommend researching the garbage collection techniques like in https://github.com/kenfox/gc-viz and seeing how we can apply it.

Deleting a file isn't the only cause where this can occur. I'm sure there may be other operations that are unscalable. Identify all of these in this issue.

### Additional context

* https://github.com/kenfox/gc-viz
* https://spin.atomicobject.com/2014/09/03/visualizing-garbage-collection-algorithms/ - discussion about the above repo

### Tasks

1. [ ] - Identify all unscalable operations and whether they apply to this situation
2. [ ] - Research the garbage collection techniques above
3. [ ] - Integrate the technique above in a lazy way so that it doesn't hog the CPU nor does it block other operations, because JS async contexts cannot be interrupted, then they must be yielded, the implementation must be as asynchronous as possible and coroutine/cooperative multitasking friendly, perhaps by the use of `setTimeout` yields (implemented as sleep promises).
4. [ ] - Consider if it would be possible to multi-thread the leveldb usage in worker threads and end up with parallel lazy GC, note that leveldb currently is not "multi-process" safe, but there is a multilevel adapter for multiprocesses, experiment with worker threads as it is using threads and not processes

CMCDragonkai:
https://github.com/MatrixAI/js-encryptedfs/pull/63#issuecomment-1102145235

Currently `ftruncate` is unscalable because it reads in all the file blocks before deleting.

[...]

PR #63, opened 2022-04-05
Upgrading @matrixai/async-init, @matrixai/async-locks, @matrixai/db, @matrixai/errors, @matrixai/workers and integrating @matrixai/resources

CMCDragonkai:
### Description

Several core libraries have been updated and js-encryptedfs needs to start using them.

1. `@matrixai/async-init` - no major changes
2. `@matrixai/async-locks` - the `INodeManager` uses a collection of `Mutex` for mutual exclusion of inodes. This means operations to individual inodes may be blocked. We can switch this with `Lock`, and in the future optimise with `RWLockWriter` or `RWLockReader`
3. `@matrixai/db` - the DB now supports proper DB transactions, we should switch the `INodeManager` API to fully support it
4. `@matrixai/errors` - we make all of our errors extend `AbstractError<T>` and also provide static descriptions to all of them, as well as use the `cause` chain
5. `@matrixai/workers` - no major changes here
6. `@matrixai/resources` - since the `@matrixai/db` no longer does any locking, the acquisition of the `DBTransaction` and `Lock` has to be done together with `withF` or `withG`

### Issues Fixed

* Related https://github.com/MatrixAI/js-polykey/issues/304
* Related https://github.com/MatrixAI/js-polykey/issues/294

### Tasks

- [x] 1. Upgrade all exceptions to use `AbstractError` and use `cause` chain and static descriptions
- [x] 2. Integrate `@matrixai/db` into `INodeManager`, remove sublevel objects and replace with full keypaths.
- [x] 3. Integrate `@matrixai/async-locks` to use the `Lock` class instead of `Mutex` from `async-mutex`
- [x] 4. Use `@matrixai/resources` `withF` and `withG` to replace some of our `transact` API.
- [x] 5. See if `FileDescriptor.write` and `FileDescriptor.read` should be using a single transaction context for the entire operation: https://github.com/MatrixAI/js-encryptedfs/pull/63#issuecomment-1094910309
- [x] 6. Removed canary check in favour of https://github.com/MatrixAI/js-db/issues/7
- [x] 7. Fix tests for transaction usage in `tests/inodes`
- [x] 8. Clean up left over commentary from migration
- ~[ ] 9. Consider optimising `EncryptedFS.ftruncate` so that it does not need to iterate over the entire file blocks before truncating, this targets #53~ - do this in #53
- [x] 10. Review `fileGetBlocks` and see if it correctly implemented - infinite loop solved by incrementing the block count inside the while loop, and added new test to `INodeManager.file.test.ts` `read sparse blocks from a file` to cover this
- [x] 11. `tests/utils.ts#expectError` has been changed to take the exception class as well, and then the `errno` object to be checked 
- [x] 12. Fixed get after delete consistency in DB and therefore in EFS, in particular `ftruncate`
- [x] 13. FD operations are mutually exclusive, this is enforced with each `FileDescriptor` having its own lock. This prevents some stream concurrency errors
- [x] 14. Reviewing all concurrency tests for correctness - some tests are too brittle, they should expect non-deterministic concurrent behaviour
  - INode creation is currently racy, it needs to be made mutually exclusive for the same name on the same directory
  - The `navigateFrom` seems racy with multiple transaction contexts being used
  - The `_open` seems racy with the final FD creation with respect to inode creation
  - The `mkdir` seems to have strange behavour when the target already exists, need to compare behaviour with real filesystem `mkdir`
- [x] 15. Updated jest, linting, typedoc dependencies

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
This is a big upgrade... it would result in `v3.5.0`, external API doesn't really change, it's mostly internal things. Lots of patterns produced now.

CMCDragonkai:
The `INodeManager.gcAll` originally used `this.db.transact` to lock up the DB every time it did any deletion operations. This should no longer be necessary with transactional iteration. No locks are even needed here since it's only called by `INodeManager.start` and `INodeManager.stop`.

CMCDragonkai:
In integrating the `@matrixai/db`. I've found the `NonEmptyArray` type to be quite annoying. It requires alot of casting.

If we remove the requirement, and that would mean `[]` has to mean something. It could mean the key `''` and thus be equivalent to `['']`.

That could simplify the code and remove all the `as unknown as KeyPath`.

CMCDragonkai:
Only issue is that some places the `keyPath` is spread into other key paths like `['data', ...keyPath]`. Prior to this happening, the an empty keypath must be turned into `['']` to ensure that a keypath always has at least 1 entry.

CMCDragonkai:
The `@matrixai/db` has been updated to `3.2.2` and now `KeyPath` `[]` becomes `['']`.

CMCDragonkai:
When I removed the `INodeManager.transact` method I noticed quite an increase in verbosity. I realised the `with*` convenience wrappers like in `RWLockWriter.withReadF()` and `RWLockWriter.withReadG()` are actually quite nice here, and they can take over the `transact` method.

However I wanted to improve the API a little bit. Since we now know how to do `NonEmptyArray`, we can also create variadic argument types that specify the last type as a callback.

So now we have:

```ts
  public async withTransactionF<T>(
    ...args: [
      ...inos: INodeIndex[],
      f: (tran: DBTransaction) => Promise<T>
    ]
  ) {
    const f = args.pop() as (tran: DBTransaction) => Promise<T>;
    return withF(
      [
        this.db.transaction(),
        this.getLocks(...args as Array<INodeIndex>)
      ],
      ([tran]) => f(tran)
    );
  }

  public withTransactionG<T, TReturn, TNext>(
    ...args: [
      ...inos: INodeIndex[],
      g: (tran: DBTransaction) => AsyncGenerator<T, TReturn, TNext>,
    ]
  ): AsyncGenerator<T, TReturn, TNext> {
    const g = args.pop() as (tran: DBTransaction) => AsyncGenerator<T, TReturn, TNext>;
    return withG(
      [
        this.db.transaction(),
        this.getLocks(...args as Array<INodeIndex>)
      ],
      ([tran]) => g(tran)
    );
  }
```

This actually conveniently replaces the `transact` methods. And you get to state up front what inodes you need to lock if any.

This is nicer than having the inodes that you need to lock at the very end.

Furthermore, notice that I "specialise" the resources to just the transaction, because the user won't have any use for the lock objects.

The user can always still compose much more complex resource contexts by using the `getLocks` and putting together the `ResourceAcquire`.

CMCDragonkai:
Also moving to selective imports so that tests can be run by themselves. 

```
import { permissions } from '@';
```

To:

```
import * as permissions from '@/permissions';
```

CMCDragonkai:
I tried getting a vscode regex to do a big find and replace. Couldn't work due to matches braces requirements.

CMCDragonkai:
There are no functions that are exposing a generator interface atm. Even streams are using the FD abstraction, and they don't actually hold lock/transaction during their lifetime. This is because every write is a single operation. So for now there's no use of `INodeManager.withTransactionG` except internally inside `INodeManager`.

CMCDragonkai:
I've found that the `FileDescriptor` methods of `write` and `read` aren't using one entire transaction for their operations. It seems like could be encapsulated in one transaction... should test this later.

CMCDragonkai:
The `EncryptedFS` has been using alot of:

```
        fd ? [fd.ino] : [],
```

And also for `target`. They should be using `fd != null` or `target != null`. The `ino` could be `0` in its type, although this is prevented by starting the resource counter at 1. I'm changing over.

CMCDragonkai:
Everything is migrated over, however I got a few test failures in `EncryptedFS`.

CMCDragonkai:
The canary check logic should be moved into DB during `await DB.create()` and `DB.start`. It doesn't make sense for it to be in the EFS. Plus now that DB has "hidden" sublevels, this can be done in the root level.

CMCDragonkai:
Current tests state, all passes except 2:

<details>

```
[nix-shell:~/Projects/js-encryptedfs]$ npm test

> encryptedfs@3.4.3 test /home/cmcdragonkai/Projects/js-encryptedfs
> jest

Determining test suites to run...
GLOBAL SETUP
 PASS  tests/workers/efsWorker.test.ts
  EFS worker
    ✓ encryption and decryption (1082 ms)
    ✓ encryption and decryption within 1 call (99 ms)

 PASS  tests/utils.test.ts
  utils
    ✓ random bytes generation (43 ms)
    ✓ key is randomly generated (4 ms)
    ✓ key generation from password is non-deterministic with random salt (766 ms)
    ✓ key generation is deterministic with a given salt (1360 ms)
    ✓ encryption and decryption (14 ms)
    ✓ block offset is position % block size (3 ms)
    ✓ number of blocks to be written
    ✓ block index start (1 ms)
    ✓ block index end (3 ms)
    ✓ block cursor (1 ms)
    ✓ buffer segmentation (6 ms)
    maybeCallback
      as a promise
        ✓ Should function
        ✓ Should throw error (18 ms)
      as a callback
        ✓ Should function (1 ms)
        ✓ Should throw error

 PASS  tests/inodes/INodeManager.test.ts (8.642 s)
  INodeManager
    ✓ inode manager is persistent across restarts (400 ms)
    ✓ transactions are locked via inodes (34 ms)
    ✓ inodes can be scheduled for deletion when there are references to them (167 ms)

 PASS  tests/fd/FileDescriptorManager.test.ts (8.91 s)
  File Descriptor Manager
    ✓ create a file descriptor manager (36 ms)
    ✓ create a file descriptor (18 ms)
    ✓ retreive a file descriptor (7 ms)
    ✓ delete a file descriptor (16 ms)
    ✓ duplicate a file descriptor (9 ms)
    ✓ read/write to fd when inode deleted from directory (210 ms)

 PASS  tests/inodes/INodeManager.symlink.test.ts (9.017 s)
  INodeManager Symlink
    ✓ create and delete symlink (274 ms)

 PASS  tests/inodes/INodeManager.file.test.ts (9.025 s)
  INodeManager File
    ✓ create a file (131 ms)
    ✓ create a file with supplied data (97 ms)
    ✓ write and read data from a file (62 ms)
    ✓ read a single block from a file (64 ms)
    ✓ write a single block from a file (41 ms)
    ✓ handle accessing blocks that the db does not have (45 ms)

 PASS  tests/inodes/INodeManager.dir.test.ts (9.339 s)
  INodeManager Directory
    ✓ create root directory (256 ms)
    ✓ create subdirectory (142 ms)
    ✓ create subdirectories (239 ms)
    ✓ delete subdirectory (106 ms)
    ✓ rename directory entry (153 ms)
    ✓ iterate directory entries (129 ms)

 PASS  tests/EncryptedFS.test.ts (9.573 s)
  EncryptedFS
    ✓ efs readiness (211 ms)
    ✓ efs is persistent across restarts (224 ms)
    ✓ creating fresh efs (192 ms)
    ✓ efs exposes constants (38 ms)
    ✓ validate key (49 ms)

 PASS  tests/fd/FileDescriptor.test.ts (9.692 s)
  File Descriptor
    ✓ create a file descriptor (65 ms)
    ✓ can set flags (20 ms)
    ✓ can set position (126 ms)
    ✓ read all the data on the file iNode (62 ms)
    ✓ read with the file descriptor at a certain position (78 ms)
    ✓ read when the return buffer length is less than the data length (66 ms)
    ✓ write to an empty file iNode (79 ms)
    ✓ overwrite a single block of a file iNode (67 ms)
    ✓ overwrite at an offset to a file iNode (75 ms)
    ✓ write past the end of a file iNode (72 ms)
    ✓ append data to the file iNode (155 ms)

 PASS  tests/EncryptedFS.streams.test.ts (10.779 s)
  EncryptedFS Streams
    readstream
      ✓ using 'for await' (335 ms)
      ✓ using 'event readable' (185 ms)
      ✓ using 'event data' (128 ms)
      ✓ respects start and end options (124 ms)
      ✓ respects the high watermark (131 ms)
      ✓ respects the start option (115 ms)
      ✓ end option is ignored without the start option (121 ms)
      ✓ can use a file descriptor (113 ms)
      ✓ with start option overrides the file descriptor position (111 ms)
      ✓ can handle errors asynchronously (36 ms)
      ✓ can compose with pipes (111 ms)
    writestream
      ✓ can compose with pipes (139 ms)
      ✓ can create and truncate files (140 ms)
      ✓ can be written into (110 ms)
      ✓ allow ignoring of the drain event, temporarily ignoring resource usage control (112 ms)
      ✓ can use the drain event to manage resource control (191 ms)
      ✓ can handle errors asynchronously (33 ms)

 PASS  tests/EncryptedFS.nav.test.ts (12.528 s)
  EncryptedFS Navigation
    ✓ EFS using callback style functions (419 ms)
    ✓ should be able to restore state (141 ms)
    ✓ should be able to navigate before root (232 ms)
    ✓ trailing slash refers to the directory instead of a file (138 ms)
    ✓ trailing slash works for non-existent directories when intending to create them (82 ms)
    ✓ trailing `/.` for mkdir should result in errors (77 ms)
    ✓ navigating invalid paths (442 ms)
    ✓ various failure situations (253 ms)
    ✓ cwd returns the absolute fully resolved path (142 ms)
    ✓ cwd still works if the current directory is deleted (106 ms)
    ✓ deleted current directory can still use . and .. for traversal (128 ms)
    ✓ can still chdir when both current and parent directories are deleted (159 ms)
    ✓ cannot chdir into a directory without execute permissions (66 ms)
    ✓ should be able to access inodes inside chroot (297 ms)
    ✓ should not be able to access inodes outside chroot (129 ms)
    ✓ should not be able to access inodes outside chroot using symlink (134 ms)
    ✓ prevents users from changing current directory above the chroot (104 ms)
    ✓ can sustain a current directory inside a chroot (70 ms)
    ✓ can chroot, and then chroot again (100 ms)
    ✓ chroot returns a running efs instance (60 ms)
    ✓ chroot start & stop does not affect other efs instances (115 ms)
    ✓ root efs instance stops all chrooted instances (121 ms)
    ✓ destroying chroot is a noop (63 ms)

 PASS  tests/EncryptedFS.perms.test.ts (14.801 s)
  EncryptedFS Permissions
    ✓ chown changes uid and gid (302 ms)
    ✓ chmod with 0 wipes out all permissions (150 ms)
    ✓ mkdir and chmod affects the mode (137 ms)
    ✓ umask is correctly applied (213 ms)
    ✓ non-root users can only chown uid if they own the file and they are chowning to themselves (213 ms)
    ✓ chmod only works if you are the owner of the file (97 ms)
    ✓ permissions are checked in stages of user, group then other (334 ms)
    ✓ permissions are checked in stages of user, group then other (using chown) (380 ms)
    ✓ --x-w-r-- permission staging (159 ms)
    ✓ file permissions --- (189 ms)
    ✓ file permissions r-- (193 ms)
    ✓ file permissions rw- (196 ms)
    ✓ file permissions rwx (187 ms)
    ✓ file permissions r-x (168 ms)
    ✓ file permissions -w- (176 ms)
    ✓ file permissions -wx (168 ms)
    ✓ file permissions --x (159 ms)
    ✓ directory permissions --- (203 ms)
    ✓ directory permissions r-- (242 ms)
    ✓ directory permissions rw- (221 ms)
    ✓ directory permissions rwx (224 ms)
    ✓ directory permissions r-x (347 ms)
    ✓ directory permissions -w- (146 ms)
    ✓ directory permissions -wx (254 ms)
    ✓ directory permissions --x (242 ms)
    ✓ permissions dont affect already opened fd (204 ms)
    ✓ chownr changes uid and gid recursively (284 ms)
    ✓ chown can change groups without any problem because we do not have a user group hierarchy (81 ms)
    ✓ --x-w-r-- do not provide read write and execute to the user due to permission staging (204 ms)

 PASS  tests/EncryptedFS.links.test.ts (15.837 s)
  EncryptedFS Links
    ✓ Symlink stat makes sense (276 ms)
    symlink
      ✓ creates symbolic links (508 ms)
      ✓ paths can contain multiple slashes (211 ms)
      ✓ can resolve 1 symlink loop (67 ms)
      ✓ can resolve 2 symlink loops (143 ms)
      ✓ can be expanded by realpath (239 ms)
      ✓ cannot be traversed by rmdir (100 ms)
      ✓ is able to be added and traversed transitively (338 ms)
      ✓ is able to traverse relative symlinks (156 ms)
      ✓ returns EACCES when a component of the 2nd name path prefix denies search permission (200 ms)
      ✓ returns EACCES if the parent directory of the file to be created denies write permission (221 ms)
      ✓ returns ELOOP if too many symbolic links were encountered in translating the name2 path name (147 ms)
      ✓ returns EEXIST if the 2nd name argument already exists as a regular (68 ms)
      ✓ returns EEXIST if the 2nd name argument already exists as a dir (51 ms)
      ✓ returns EEXIST if the 2nd name argument already exists as a block (61 ms)
      ✓ returns EEXIST if the 2nd name argument already exists as a symlink (54 ms)
    unlink
      ✓ can remove a link to a regular (81 ms)
      ✓ can remove a link to a block (55 ms)
      ✓ successful updates ctime of a regular (122 ms)
      ✓ successful updates ctime of a block (122 ms)
      ✓ unsuccessful does not update ctime of a regular (111 ms)
      ✓ unsuccessful does not update ctime of a block (111 ms)
      ✓ does not traverse symlinks (198 ms)
      ✓ returns ENOTDIR if a component of the path prefix is not a directory (113 ms)
      ✓ returns ENOENT if the named file does not exist (78 ms)
      ✓ returns EACCES when search permission is denied for a component of the path prefix (122 ms)
      ✓ returns EACCES when write permission is denied on the directory containing the link to be removed (119 ms)
      ✓ returns ELOOP if too many symbolic links were encountered in translating the pathname (115 ms)
      ✓ returns EISDIR if the named file is a directory (65 ms)
      ✓ will not immeadiately free a file (172 ms)
    link
      ✓ creates hardlinks to regular (214 ms)
      ✓ creates hardlinks to block (243 ms)
      ✓ successful updates ctime of regular (118 ms)
      ✓ successful updates ctime of block (108 ms)
      ✓ unsuccessful does not update ctime of regular (123 ms)
      ✓ unsuccessful does not update ctime of block (121 ms)
      ✓ should not create hardlinks to directories (50 ms)
      ✓ can create multiple hardlinks to the same file (151 ms)
      ✓ returns ENOTDIR if a component of either path prefix is a regular (147 ms)
      ✓ returns ENOTDIR if a component of either path prefix is a dir (22 ms)
      ✓ returns ENOTDIR if a component of either path prefix is a block (120 ms)
      ✓ returns ENOTDIR if a component of either path prefix is a symlink (18 ms)
      ✓ returns EACCES when a component of either path prefix denies search permission (212 ms)
      ✓ returns EACCES when the requested link requires writing in a directory with a mode that denies write permission (199 ms)
      ✓ returns ELOOP if too many symbolic links were encountered in translating one of the pathnames (151 ms)
      ✓ returns ENOENT if the source file does not exist (103 ms)
      ✓ returns EEXIST if the destination regular does exist (110 ms)
      ✓ returns EEXIST if the destination dir does exist (94 ms)
      ✓ returns EEXIST if the destination block does exist (76 ms)
      ✓ returns EEXIST if the destination symlink does exist (79 ms)
      ✓ returns EPERM if the source file is a directory (86 ms)

 FAIL  tests/EncryptedFS.files.test.ts (16.472 s)
  EncryptedFS Files
    ✓ File stat makes sense (331 ms)
    ✓ Uint8Array data support (242 ms)
    ✓ URL path support (240 ms)
    lseek
      ✓ can seek different parts of a file (134 ms)
      ✓ can seek beyond the file length and create a zeroed "sparse" file (143 ms)
    fallocate
      ✓ can extend the file length (94 ms)
      ✓ does not touch existing data (96 ms)
      ✓ will only change ctime (120 ms)
    truncate
      ✓ will change mtime and ctime (181 ms)
      ✕ truncates the fd position (204 ms)
    read
      ✓ can be called using different styles (126 ms)
      ✓ file can be called using different styles (132 ms)
      ✓ file moves with fd position (132 ms)
      ✓ moves with the fd position (104 ms)
      ✓ does not change fd position according to position parameter (313 ms)
    write
      ✓ can be called using different styles (104 ms)
      ✓ file can be called using different styles (259 ms)
      ✓ moves with the fd position (113 ms)
      ✓ can make 100 files (1760 ms)
      ✓ does not change fd position according to position parameter (97 ms)
      ✓ respects the mode (138 ms)
      ✓ file writes from the beginning, and does not move the fd position (91 ms)
      ✓ with O_APPEND always set their fd position to the end (159 ms)
      ✓ can copy files (152 ms)
      ✓ using append moves with the fd position (115 ms)
    open
      ✓ opens a file if O_CREAT is specified and the file doesn't exist (148 ms)
      ✓ updates parent directory ctime/mtime if file didn't exist (134 ms)
      ✓ doesn't update parent directory ctime/mtime if file existed (142 ms)
      ✓ returns ENOTDIR if a component of the path prefix is a regular (64 ms)
      ✓ returns ENOTDIR if a component of the path prefix is a block (60 ms)
      ✓ returns ENOENT if a component of the path name that must exist does not exist or O_CREAT is not set and the named file does not exist (88 ms)
      ✓ returns EACCES when search permission is denied for a component of the path prefix (145 ms)
      ✓ returns EACCES when the required permissions are denied for a regular file (476 ms)
      ✓ returns EACCES when the required permissions are denied for adirectory (290 ms)
      ✓ returns EACCES when O_TRUNC is specified and write permission is denied (210 ms)
      ✓ returns ELOOP if too many symbolic links were encountered in translating the pathname (84 ms)
      ✓ returns EISDIR when trying to open a directory for writing (71 ms)
      ✓ returns ELOOP when O_NOFOLLOW was specified and the target is a symbolic link (60 ms)
      ✓ returns EEXIST when O_CREAT and O_EXCL were specified and the regular exists (90 ms)
      ✓ returns EEXIST when O_CREAT and O_EXCL were specified and the dir exists (72 ms)
      ✓ returns EEXIST when O_CREAT and O_EXCL were specified and the block exists (78 ms)
      ✓ returns EEXIST when O_CREAT and O_EXCL were specified and the symlink exists (83 ms)

  ● EncryptedFS Files › truncate › truncates the fd position

    expect(received).toEqual(expected) // deep equality

    - Expected  - 2
    + Received  + 2

      Object {
        "data": Array [
          100,
    -     98,
    -     99,
    +     101,
    +     102,
        ],
        "type": "Buffer",
      }

      183 |       await efs.ftruncate(fd, 4);
      184 |       await efs.read(fd, buf, 0, buf.length);
    > 185 |       expect(buf).toEqual(Buffer.from('dbc'));
          |                   ^
      186 |       await efs.close(fd);
      187 |     });
      188 |   });

      at Object.<anonymous> (tests/EncryptedFS.files.test.ts:185:19)

 PASS  tests/EncryptedFS.dirs.test.ts (17.557 s)
  EncryptedFS Directories
    ✓ Directory stat makes sense (319 ms)
    ✓ Empty root directory at startup (102 ms)
    file descriptors
      ✓ can change stats, permissions and flush data (215 ms)
      ✓ cannot perform read or write operations (186 ms)
      ✓ inode nlink becomes 0 after deletion of the directory (151 ms)
    rmdir
      ✓ should be able to remove directories (390 ms)
      ✓ cannot delete current directory using . (88 ms)
      ✓ cannot delete parent directory using .. even when current directory is deleted (174 ms)
      ✓ cannot create inodes within a deleted current directory (197 ms)
      ✓ returns ENOENT if the named directory does not exist (04) (95 ms)
      ✓ returns ELOOP if too many symbolic links were encountered in translating the pathname (175 ms)
      ✓ returns ENOTEMPTY if the named directory contains files other than '.' and '..' in it for regular (144 ms)
      ✓ returns ENOTEMPTY if the named directory contains files other than '.' and '..' in it for dir (106 ms)
      ✓ returns ENOTEMPTY if the named directory contains files other than '.' and '..' in it for block (115 ms)
      ✓ returns ENOTEMPTY if the named directory contains files other than '.' and '..' in it for symlink (109 ms)
      ✓ returns EACCES when search permission is denied for a component of the path prefix (213 ms)
      ✓ returns EACCES when write permission is denied on the directory containing the link to be removed (187 ms)
      ✓ recursively deletes the directory if it contains regular (135 ms)
      ✓ recursively deletes the directory if it contains dir (122 ms)
      ✓ recursively deletes the directory if it contains block (128 ms)
      ✓ recursively deletes the directory if it contains symlink (161 ms)
      ✓ recursively deletes a deep directory (684 ms)
    mkdir & mkdtemp
      ✓ is able to make directories (328 ms)
      ✓ can create temporary directories (114 ms)
      ✓ should not make the root directory (24 ms)
      ✓ trailing '/.' should not result in any errors (201 ms)
      ✓ returns EACCES when write permission is denied on the parent directory of the directory to be created (183 ms)
      ✓ returns EEXIST if the named regular exists (93 ms)
      ✓ returns EEXIST if the named dir exists (92 ms)
      ✓ returns EEXIST if the named block exists (77 ms)
      ✓ returns EEXIST if the named symlink exists (88 ms)
    rename
      ✓ can rename a directory (80 ms)
      ✓ cannot rename the current or parent directory to a subdirectory (117 ms)
      ✓ cannot rename where the old path is a strict prefix of the new path (166 ms)
      ✓ changes name but inode remains the same for regular (150 ms)
      ✓ changes name but inode remains the same for block (140 ms)
      ✓ changes name for dir (81 ms)
      ✓ changes name for regular file (135 ms)
      ✓ unsuccessful of regular does not update ctime (92 ms)
      ✓ unsuccessful of dir does not update ctime (86 ms)
      ✓ unsuccessful of block does not update ctime (76 ms)
      ✓ unsuccessful of symlink does not update ctime (76 ms)
      ✓ returns ENOENT if a component of the 'from' path does not exist, or a path prefix of 'to' does not exist (92 ms)
      ✓ returns EACCES when a component of either path prefix denies search permission (224 ms)
      ✓ returns EACCES when the requested link requires writing in a directory with a mode that denies write permission (220 ms)
      ✓ returns ELOOP if too many symbolic links were encountered in translating one of the pathnames (155 ms)
      ✓ returns ENOTDIR if a component of either path prefix is a regular (142 ms)
      ✓ returns ENOTDIR if a component of either path prefix is a block (117 ms)
      ✓ returns ENOTDIR when the 'from' argument is a directory, but 'to' is a regular (80 ms)
      ✓ returns ENOTDIR when the 'from' argument is a directory, but 'to' is a block (88 ms)
      ✓ returns ENOTDIR when the 'from' argument is a directory, but 'to' is a symlink (84 ms)
      ✓ returns EISDIR when the 'to' argument is a directory, but 'from' is a regular (86 ms)
      ✓ returns EISDIR when the 'to' argument is a directory, but 'from' is a block (76 ms)
      ✓ returns EISDIR when the 'to' argument is a directory, but 'from' is a symlink (86 ms)
      ✓ returns EINVAL when the 'from' argument is a parent directory of 'to' (93 ms)
      ✓ returns ENOTEMPTY if the 'to' argument is a directory and contains regular (107 ms)
      ✓ returns ENOTEMPTY if the 'to' argument is a directory and contains dir (102 ms)
      ✓ returns ENOTEMPTY if the 'to' argument is a directory and contains block (96 ms)
      ✓ returns ENOTEMPTY if the 'to' argument is a directory and contains symlink (97 ms)
      ✓ changes file ctime for regular (103 ms)
      ✓ changes file ctime for dir (84 ms)
      ✓ changes file ctime for block (77 ms)
      ✓ changes file ctime for symlink (80 ms)
      ✓ succeeds when destination regular is multiply linked (142 ms)
      ✓ succeeds when destination block is multiply linked (115 ms)

 FAIL  tests/EncryptedFS.concurrent.test.ts (24.681 s)
  EncryptedFS Concurrency
    ✓ Renaming a directory at the same time with two different calls (355 ms)
    ✓ Reading a directory while adding/removing entries in the directory (246 ms)
    ✓ Reading a directory while removing the directory (277 ms)
    ✓ Reading a directory while renaming entries (278 ms)
    ✓ File metadata changes while reading/writing a file. (109 ms)
    ✓ Dir metadata changes while reading/writing a file. (177 ms)
    ✓ Read stream and write stream to same file (1158 ms)
    ✓ One write stream and one fd writing to the same file (88 ms)
    ✓ One read stream and one fd writing to the same file (176 ms)
    ✓ One write stream and one fd reading to the same file (78 ms)
    ✓ One read stream and one fd reading to the same file (168 ms)
    ✓ Two write streams to the same file (2097 ms)
    ✓ Writing a file and deleting the file at the same time using writeFile (68 ms)
    ✓ opening a file and deleting the file at the same time (71 ms)
    ✓ Writing a file and deleting the file at the same time for fd (148 ms)
    ✓ Writing a file and deleting the file at the same time for stream (125 ms)
    ✓ Appending to a file that is being written to for fd  (1130 ms)
    ✕ Appending to a file that is being written for stream (86 ms)
    ✓ Copying a file that is being written to for fd (192 ms)
    ✓ Copying a file that is being written to for stream (284 ms)
    ✓ removing a dir while renaming it. (120 ms)
    concurrent file writes
      ✓ 10 short writes with efs.writeFile. (508 ms)
      ✓ 10 long writes with efs.writeFile. (2155 ms)
      ✓ 10 short writes with efs.write. (179 ms)
      ✓ 10 long writes with efs.write. (4299 ms)
    Allocating/truncating a file while writing (stream or fd)
      ✓ Allocating while writing to fd (93 ms)
      ✓ Truncating while writing to fd (157 ms)
      ✓ Allocating while writing to stream (100 ms)
      ✓ Truncating while writing to stream (115 ms)
    Changing fd location in a file (lseek) while writing/reading (and updating) fd pos
      ✓ Seeking while writing to file. (62 ms)
      ✓ Seeking while reading a file. (64 ms)
      ✓ Seeking while updating fd pos. (53 ms)
    checking if nlinks gets clobbered.
      ✓ when creating and removing the file. (215 ms)
      ✓ when creating and removing links. (251 ms)

  ● EncryptedFS Concurrency › Appending to a file that is being written for stream

    expect(received).toContain(expected) // indexOf

    Expected substring: "A"
    Received string:    "BBBBBBBBBB"

      820 | 
      821 |     const fileContents = (await efs.readFile('file')).toString();
    > 822 |     expect(fileContents).toContain('A');
          |                          ^
      823 |     expect(fileContents).toContain('B');
      824 |     expect(fileContents).toContain('AB');
      825 | 

      at Object.<anonymous> (tests/EncryptedFS.concurrent.test.ts:822:26)

Test Suites: 2 failed, 14 passed, 16 total
Tests:       2 failed, 314 passed, 316 total
Snapshots:   0 total
Time:        25.165 s, estimated 26 s
Ran all test suites.
GLOBAL TEARDOWN
npm ERR! Test failed.  See above for more details.
```

</details>

First:

```

  ● EncryptedFS Files › truncate › truncates the fd position

    expect(received).toEqual(expected) // deep equality

    - Expected  - 2
    + Received  + 2

      Object {
        "data": Array [
          100,
    -     98,
    -     99,
    +     101,
    +     102,
        ],
        "type": "Buffer",
      }

      183 |       await efs.ftruncate(fd, 4);
      184 |       await efs.read(fd, buf, 0, buf.length);
    > 185 |       expect(buf).toEqual(Buffer.from('dbc'));
          |                   ^
      186 |       await efs.close(fd);
      187 |     });
      188 |   });

      at Object.<anonymous> (tests/EncryptedFS.files.test.ts:185:19)
```

```
  ● EncryptedFS Concurrency › Appending to a file that is being written for stream

    expect(received).toContain(expected) // indexOf

    Expected substring: "A"
    Received string:    "BBBBBBBBBB"

      820 | 
      821 |     const fileContents = (await efs.readFile('file')).toString();
    > 822 |     expect(fileContents).toContain('A');
          |                          ^
      823 |     expect(fileContents).toContain('B');
      824 |     expect(fileContents).toContain('AB');
      825 | 

      at Object.<anonymous> (tests/EncryptedFS.concurrent.test.ts:822:26)
```

Both appear to involve concurrent writing/reading to file blocks which is the trickiest part, and that which is impacted by https://github.com/MatrixAI/js-encryptedfs/pull/63#issuecomment-1094910309

CMCDragonkai:
Replacing `INodeManager`'s lock collection with `LockBox` from `@matrixai/async-locks` 2.2.0.

CMCDragonkai:
No problems integrating `LockBox`.

In doing so I discovered that there's no need to remove entries from the lock collection since the lock entries are now removed when there is nothing holding them.

This does mean that it has to construct the lock object each time someone is trying to lock it. Which can be a bit inefficient.

But at the same time, not having to keep it around reduces programmer burden to figure out when to clear or delete a lock entry.

In the future, the `LockBox` can additionally be cached if necessary.

CMCDragonkai:
For the first test failure involving `truncates the fd position`.

It appears the `efs.ftruncate(fd, 4);` isn't doing anything.

The data in teh `/fdtest` should be `abcdef`, and when truncating it should become `abcd`. The pointer should be at `d`, and therefore the end result should be `dbc`.

I confirmed this with a normal `fs`. However this is now broken. Need to trace through `ftruncate`.

CMCDragonkai:
The `ftruncate` leads to the `INodeManager.fileWriteBlock`. At the very end, it says it's writing `4` bytes. However the resulting block is still `abcdef` and not `abcd`.

Need to compare with what is in `master`.

CMCDragonkai:
In master, it doesn't go into that case at all. It instead goes to `fileWriteBlock`:

```
    if (!block) {
      await tran.put(dataDomain, key, data, true);
      bytesWritten = data.length;
    } else {
```

The reason is because:

```
    let block = await this.fileGetBlock(tran, ino, idx);
```

Is `undefined`. It is looking up `ino = 2` and `idx = 0`.

This is quite strange. Because in `master`, the block doesn't exist, and thus `fileGetBlock` is undefined.

However in the current PR, it does exist.

Now should it exist or not? If you already have a file that has `abcdef`. If you open up a `r+` descriptor, that doesn't clear the file.

Therefore it seems to make sense that the block would still exist, and the `master` is wrong to not have the buffer.

CMCDragonkai:
Reduced this problem to the simplest reproduction:

```ts
import fs from 'fs/promises';

async function main () {

  await fs.writeFile('./tmp/fdtest', 'abcdef');
  const fd = await fs.open('./tmp/fdtest', 'r+');
  await fd.truncate(4);
  await fd.close();
  // File is abcd
  console.log(await fs.readFile('./tmp/fdtest', { encoding: 'utf-8' }));

}

main();
```

vs:

```ts
import os from 'os';
import fs from 'fs';
import pathNode from 'path';
import Logger, { StreamHandler, LogLevel } from '@matrixai/logger';
import { EncryptedFS, utils } from './src';

async function main () {
  const logger = new Logger('EncryptedFS Files', LogLevel.WARN, [
    new StreamHandler(),
  ]);
  const dataDir = await fs.promises.mkdtemp(
    pathNode.join(os.tmpdir(), 'encryptedfs-test-'),
  );
  const dbPath = `${dataDir}/db`;
  const dbKey: Buffer = utils.generateKeySync(256);
  const efs = await EncryptedFS.createEncryptedFS({
    dbKey,
    dbPath,
    umask: 0o022,
    logger,
  });

  await efs.writeFile('/fdtest', 'abcdef');

  // File is abcdef
  console.log(await efs.readFile('/fdtest', { encoding: 'utf-8'}));

  const fd = await efs.open('/fdtest', 'r+');

  // File is abcdef
  console.log(await efs.readFile('/fdtest', { encoding: 'utf-8'}));

  await efs.ftruncate(fd, 4);
  await efs.close(fd);

  // File should be abcd, but is instead abcdef
  console.log(await efs.readFile('/fdtest', { encoding: 'utf-8'}));

}

main();
```

In the current branch, the resulting file isn't truncated. It's still `abcdef`.

In the `master` branch it does it correctly to `abcd` just like the normal `fs`.

CMCDragonkai:
The main difference is in `fileWriteBlock`.

When looking up the block during truncation, in `master`, we have `undefined`, which means it ends up writing the `abcd`.

However in the current branch, it does find the block which means it doesn't write a new block being just `abcd`, but instead it would "write-over" the `abcdef`, and thus end up with the same result.

The question is why does `master` see no block while this current branch does see the block.

CMCDragonkai:
The `fileGetBlocks` has been changed from `master` using `createReadStream`, to now using the `tran.iterator` in this current branch.

This could mean that it is correctly acquiring the block in the current branch, because it is still `abcdef`.

I'm not sure if the `fileGetBlocks` is correct anyway since it has a while loop that as long as the buffer index is not equal to the block count, it yields empty blocks. But this while loop will never become true/false...?

CMCDragonkai:
I noticed that `ftruncate` has to call `await this.iNodeMgr.fileClearData(fd.ino, tran);` prior to calling `fileSetBlocks`.

This would mean that it is intended that the block is `undefined` if `fileClearData` is meant to clear all the blocks in that sublevel.

However it appears that even though it is clearing the right key: `[ 'INodeManager', 'data', '2', <Buffer 00> ]`, the block is still being found afterwards. This might be a bug in our branch then.

CMCDragonkai:
I've checked, right after calling `await tran.del([...dataPath, k])` I try to  then `await tran.get([...dataPath, k]);` it appears the `await tran.del` has no effect. Possibly due to the `tran.iterator`.

I will need to investigate the `js-db` for a possible bug here.

CMCDragonkai:
Confirmed this is a problem in `js-db`.

```ts
import Logger, { LogLevel, StreamHandler } from '@matrixai/logger';
import { DB } from './src';
import * as testUtils from './tests/utils';

async function main () {

  const logger = new Logger(`${DB.name} Test`, LogLevel.WARN, [
    new StreamHandler(),
  ]);

  const crypto = {
    key: testUtils.generateKeySync(256),
    ops: {
      encrypt: testUtils.encrypt,
      decrypt: testUtils.decrypt,
    },
  };

  const db = await DB.createDB({
    dbPath: './tmp/db',
    crypto,
    fresh: true,
    logger
  });

  const acquireTran = db.transaction();

  await db.put(
    ['INodeManager', 'data', '2', Buffer.from([0])],
    Buffer.from('abcdef'),
    true
  );

  let releaseTran, tran;

  [releaseTran, tran] = await acquireTran();

  console.log('BEFORE DELETE', await tran!.get(
    ['INodeManager', 'data', '2', Buffer.from([0])],
    true
  ));

  await tran!.del(
    ['INodeManager', 'data', '2', Buffer.from([0])],
  );

  console.log('AFTER DELETE', await tran!.get(
    ['INodeManager', 'data', '2', Buffer.from([0])],
    true
  ));

  await releaseTran();

  console.log('AFTER TRANSACTION', await db.get(
    ['INodeManager', 'data', '2', Buffer.from([0])],
    true
  ));

  await db.stop();

}

main();
```

This shows that when there's data in the database at a particular key. If I later start a transaction, acquire that data, but then delete it, if I get it again, this doesn't reflect the mutation I had just done. It thinks that data is still there.

Only after the transaction is committed, is the data fully removed. Which is what we want, but at the very least within the transaction we need to have some consistency in that within a transaction get after delete should be `undefined`.

CMCDragonkai:
Additionally some improvements on DB is possible here:

* DB should have convenience methods `withTransactionF` and `withTransactionG`.
* The `ResourceAcquire` type right now always specifies that the resource is optional. However that should only the default generic parameter, if it is specified, then the resource should always be there. That avoids having to use `tran!` when we know it's there.

CMCDragonkai:
Upgrading to `@matrixai/db` to 3.3.0 has fixed this bug of get after delete consistency.

Furthermore I reckon the `ftruncate` isn't entirely efficient here. Right now it loads all the blocks into memory, and then proceeds to overwrite the existing blocks in the DB.

With our transactional iteration, we should be able to identify the block index we need to truncate to, and then do a more precise deletion of the blocks above, and the overwrite of that specific block that needs to be done. Will address this after all bugs are solved.

However the second bug is still a problem.

```
  ● EncryptedFS Concurrency › Appending to a file that is being written for stream

    expect(received).toContain(expected) // indexOf

    Expected substring: "A"
    Received string:    "BBBBBBBBBB"

      820 | 
      821 |     const fileContents = (await efs.readFile('file')).toString();
    > 822 |     expect(fileContents).toContain('A');
```

CMCDragonkai:
Just discovered that `createWriteStream` and `createReadStream` seems to be incorrect. It's written asynchronously, but it is unnecessary to make these asynchronous as they just create a stream object.

CMCDragonkai:
The stream test failing because if we have 2 concurrent operations, one that writes from a stream and another is appending to the file, we should see something like:

```
AAAAAAAAAABBBBBBBBBB
```

or

```
BBBBBBBBBBAAAAAAAAAA
```

However in our EFS we are currently only seeing: `BBBBBBBBBB`.

This probably has to do with our new `DBTransaction`.

```
import fs from 'fs';
import fsPromises from 'fs/promises';

async function main () {

  await fsPromises.writeFile('./tmp/streamtest', '');

  const writeStream = fs.createWriteStream('/tmp/streamtest');

  await Promise.all([
    new Promise((resolve) => {
      writeStream.write(Buffer.from('AAAAAAAAAA'), () => {
        writeStream.end(() => {
          resolve(null);
        });
      });
    }),
    fsPromises.appendFile('/tmp/streamtest', 'BBBBBBBBBB'),
  ]);

  // Shows AAAAAAAAAABBBBBBBBBB
  console.log((await fsPromises.readFile('/tmp/streamtest')).toString());

}

main();
```

CMCDragonkai:
This is easily reproduced with just:

```ts
  await efs.writeFile('/fdtest', '');

  await Promise.all([
    efs.appendFile('/fdtest', 'AAA'),
    efs.appendFile('/fdtest', 'BBB'),
  ]);

  console.log(await efs.readFile('/fdtest', { encoding: 'utf-8' }));

  await efs.stop();
```

I think this might be a lost update problem.

Basically 2 transaction are committing to the same key. However they should be locking onto the same key and therefore the appends should end up being written one at a time.

CMCDragonkai:
This may be related to task 5, that is a single transaction context for the entire write/read operation rather than how it is done now.

CMCDragonkai:
I'm thinking that each `FileDescriptor` itself should have a lock. This is because `FileDescriptor` has the position state that is being mutated during read and write operations. It would make sense, that each individual read/write operations should be blocking each other.

Checking this https://stackoverflow.com/questions/5057737/simultaneous-read-on-file-descriptor-from-two-threads it seems POSIX is unspecified on this operation.

I need to check with normal fs, and see how concurrent `read`, or `write` works here.

A solution may turn out to be:

1. Having a lock on each `FileDescriptor` instance to synchronise the `read` and `write`, `setPos` methods.
2. Also ensuring that the transaction context covers the entire `read`, `write` and `setPos` operation.

CMCDragonkai:
This script proves that at least for JS, concurrent reads and writes on a file descriptor is synchronised.

```ts
import fs from 'fs';
import fsPromises from 'fs/promises';

async function main () {

  await fsPromises.writeFile('./tmp/ctest', 'ABCDEFG');

  let fd = fs.openSync('./tmp/ctest', 'r');

  const b1 = Buffer.alloc(5);
  const b2 = Buffer.alloc(5);
  console.log(b1, b2);

  const results = await Promise.all([
    new Promise((resolve) => {
      // @ts-ignore
      fs.read(fd, { buffer: b1 }, (e, bytesRead, buf) => {
        console.log(e, bytesRead, buf);
        resolve(buf);
      });
    }),
    new Promise((resolve) => {
      // @ts-ignore
      fs.read(fd, { buffer: b2 }, (e, bytesRead, buf) => {
        console.log(e, bytesRead, buf);
        resolve(buf);
      });
    }),
  ]);

  fs.closeSync(fd);
  fd = fs.openSync('./tmp/ctest', 'w');

  // @ts-ignore
  console.log(results.map(b => b.toString()));

  await Promise.all([
    new Promise<void>((resolve) => {
      // @ts-ignore
      fs.write(fd, Buffer.from('aaa'), () => {
        resolve();
      });
    }),
    new Promise<void>((resolve) => {
      fs.write(fd, Buffer.from('bbb'), () => {
        resolve();
      });
    }),
  ]);

  console.log(await fsPromises.readFile('./tmp/ctest', { encoding: 'utf-8' }));

}

main();
```

CMCDragonkai:
Since `setPos` has an optional `tran` parameter. I'm also adding the optional `tran` parameter to `read` and `write`. Although not all EFS methods would actually end up using it.

CMCDragonkai:
In applying the lock and a single transaction context in task 5, we now fixed that specific test failure, however many other tests in the `tests/EncryptedFS.concurrent.test.ts` are broken. I believe this was due to an incorrect implementation done for the file descriptor from the beginning, so the tests for the file descriptors are themselves incorrect.

Need to review all the concurrent tests again. @tegefaulkes

CMCDragonkai:
Some of the concurrent tests are quite brittle as they expect deterministic concurrent behaviour. This is incorrect, for example when reading a directory and deleting the directory, it should deal with the possibility that the directory is first deleted, or last deleted.

I'm simplifying these tests so that they actually test for non-determinism.

CMCDragonkai:
Another concurrency bug this time by doing 10 append file operations on a file that doesn't exist.

The problem here is that because the file doesn't exist, each append file operation is trying to create a new file, and therefore a new inode number.

Our locks on inode doesn't help here because it's all locking on different inode numbers.

If the file already did exist, then concurrent `appendFile` would work as they would all lock on the same inode number.

It seems what we need to do is to ensure that our open on the same "key" here is the same lock. In this case the key is the directory entry for that directory.

The same situation is actually happening with concurrent `writeFile` when the file doesn't exist, because each operation is technically creating new inodes.

See this:

```ts
    test.only('10 short writes with EncryptedFS.writeFile', async () => {
      const contents = [
        'one',
        'two',
        'three',
        'four',
        'five',
        'six',
        'seven',
        'eight',
        'nine',
        'ten',
      ];
      const promises: Array<any> = [];
      for (const content of contents) {
        promises.push(efs.writeFile('test', content));
      }
      await Promise.all(promises);
      // One of the writes wins
      expect(contents).toContainEqual(await efs.readFile('test', { encoding: 'utf-8' }));

      // Only ten or inode 11 is supposed to be left
      console.log(await efs.readFile('test', { encoding: 'utf-8' }));
      // However all of these inodes still exist (they are just gc-able)
      for (const i of [2,3,4,5,6,7,8,9,10,11]) {
        console.log(await iNodeMgr.get(i as INodeIndex));
      }
    });
```

This is basically the object-map locking problem (https://gist.github.com/CMCDragonkai/f58f08e7eaab0430ed4467ca35527a42). How do we ensure that the creation of a specific object is also locked.

When I developed the `LockBox` it was intended to also help this particular usecase, so we need to figure out how to resolve this in EFS, and it can be used in the rest of PK.

CMCDragonkai:
The problem is that our lockbox is mapping inode index to locks, but at the abstraction level of EFS, it's trying to create the same file indexed by the file name/directory entry.

So while it is correct that we should be indexing locks by the inode index, we also need to index the inode index by the directory entry/file name.

This would imply that each directory should be locking the inode allocation for a given entry as well. So if there 10 async requests for a new inode for the entry `file`, then each request should be given the same inode index.

CMCDragonkai:
There are potential solutions for this.

1. Introduce a `LockBox` for each directory that is indexed on each directory entry. Remember `LockBox` are in-memory objects, and we no longer have a `Directory` object. It's all managed by `INodeManager`. Which means keeping references to the `LockBox` would have be done by the `INodeManager` which is kind of annoying. You'd have to create a `LockBox` for each directory inode possibly dynamically. This would enforce mutual exclusion at the directory-level so that creation operations involving the same entry can be blocked.
2. Coalesce inode allocation for each directory such that allocating an inode for entry X, always results in Y. That is for the same X, get the same Y. Once you have the same Y which is the same inode index, then you can rely on the mutual exclusion already maintained by the `INodeManager` lockbox indexed by inode index. This can be done without introducing additional locks simply because inode allocation is synchronous operation on the `js-resource-counter` tree structure.

I prefer solution 2, and will be attempting this. This however requires an intermediate in-memory map data structure that maps directory entry names to allocated inode indexes. This would have to be maintained by `INodeManager` or `EncryptedFS`. It seems like a `INodeManager` problem. Because `Directory` objects do not exist, you still have to dynamically (synchronously) construct this map structure for when directory inodes are being interacted with. At the very least, solution 2 doesn't require using the `LockBox`, because inode allocations are synchronous.

CMCDragonkai:
It seems we might need `tran.queueFinally` to be able to call `inoAllocated(navigated)` so that way the intermediate entries can be removed.

```
  /**
   * Allocates an INodeIndex
   * This is a purely functional method
   * If passed without the navigated parameter, it assumed that this is for the root INodeIndex
   * If passed with the navigated parameter, it will return the same INodeIndex for the same input parameters
   * If the inode allocation is unsuccessful, you must deallocate the inode
   * Not doing so will result in memory-leak of INodeIndexes
   */
  public inoAllocate(navigated?: Readonly<{ dir: INodeIndex, name: string }>): INodeIndex {
    let key: string;
    if (navigated == null) {
      key = ''
    } else {
      key = navigated.dir + navigated.name;
    }
    let iNodeIndex = this.iNodeAllocations.get(key);
    if (iNodeIndex == null) {
      iNodeIndex = this.iNodeCounter.allocate();
      this.iNodeAllocations.set(key, iNodeIndex!);
    }
    return iNodeIndex!;
    // return this.iNodeCounter.allocate();
  }

  /**
   * Remove the entry from the inodeAllocations
   * If you fail to do so, you should also delete it from the table
   */
  public inoAllocated(navigated?: Readonly<{ dir: INodeIndex, name: string }>): void {
    let key: string;
    if (navigated == null) {
      key = ''
    } else {
      key = navigated.dir + navigated.name;
    }
    this.iNodeAllocations.delete(key);
  }

  public inoDeallocate(ino: INodeIndex): void {
    return this.iNodeCounter.deallocate(ino);
  }
```

Notice that `inoDeallocate` is not involved here. Instead there is a `inoAllocated` that finishes by removing the mapping entry from the intermediate data structure.

However this exposes another bug, where the file no longer exists if there are multiple concurrent `writeFile`.

It doesn't quite make sense, because `readdir` is showing that the entry does exist.

This may be because `navigatedFrom` is also breaking up the transaction contexts, and we really need that to be within 1 transaction... I'm not sure if there was a good reason to break up the transaction contexts.

CMCDragonkai:
Will need to review https://gitlab.com/MatrixAI/Engineering/Polykey/js-encryptedfs/-/merge_requests/46#note_649228698 for reasons for why the transaction contexts were broken up in `navigateFrom`.

CMCDragonkai:
I'm thinking the `inoAllocate` and `inoDeallocate` can be part of the `ResourceAcquire` type as well. This way we can compose the inode allocation with lock allocation. Right now, we usually want to lock a given inode, but if we are creating a new inode (on specific name in a particular directory), then we also want to lock on that. I think there might be a way to incorporate this into the `withTransactionF`, by passing something like an `undefined` to represent the need for a new inode... This can avoid us having to use `tran.queueFailure` and `tran.queueSuccess`.

The main thing is that concurrent operations get the same inode when trying to create the same key. Afterwards, once the inode is created, the entry can removed, as all subsequent operations will find that inode and should not be trying to create the inode.

CMCDragonkai:
Turns out the `fileGetBlocks` was buggy. It didn't handle the case of sparse blocks, where the file descriptor is set to a position beyond the end of the file. When returning the zeroed-out blocks, it's important to increment the `blockCount` otherwise it will result in an infinite loop.

I added a new test `read sparse blocks from a file` to `tests/inodes/INodeManager.file.test.ts` to cover this case.

Note that `fallocate` and `ftruncate` could create sparse files, but at the moment, the implementation doesn't do this. I believe it just writes zeroed-out blocks, but I haven't looked too deeply.

CMCDragonkai:
At some point this the `O_TRUNC` was added in:

```
flags & (vfs.constants.O_WRONLY | (flags & (vfs.constants.O_RDWR | (flags & vfs.constants.O_TRUNC))))
```

From originally:

```
            flags & (vfs.constants.O_WRONLY | (flags & vfs.constants.O_RDWR))
```

I believe this was a linting change. I'm not sure if it has the same effect.

CMCDragonkai:
I have some success by restructuring the `_open`.

Note that `INodeManager.inoAllocation()` is now a method that returns `ResourceAcquire<INodeIndex>`. This enables futher composition to become `withF`.

Furthermore although `ResourceAcquire` is asynchronous. It is actually synchronous internally, there's no await internally. This means it really behaves synchronously because promises are eagerly-evaluated and not lazy-evaluated.

For example:

```
function doThisSync () {
  console.log('called do this sync');
  return 1;
}

async function doThis () {
  console.log('called do this async');
  return 1;
}
async function main () {
  const p = doThis();
  doThisSync();
  await p;
}

main();
```

Running this will show `called do this async` gets called first before `called do this sync`. In fact, anything synchronous before the first await in an asynchronous function would be synchronously executed.

This should mean that our mapping to inode indexes should work.

However we then need to check if the inode did in fact get created, and if so, we need to restart the opening procedure. I've done this by using a while loop. But it does involving releasing the transaction and going back to attempting to lock the transaction. This can mean things get out of order since we aren't preserving the same transaction. I'd like to preserve it, but this requires some thinking, but it may just be cosmetic. Further testing required.

CMCDragonkai:
Additional issue is that the `navigateFrom` uses alot of transactional contexts. When I changed to bringing them into 1 transactional context, it resulted in not being able to find a particular target. A bit strange. So testing this.

CMCDragonkai:
And now that we have switched over `_open` to use `inoAllocation`, the other methods of `EncryptedFS` that creates inodes will need to do the same such as `mkdir`, `mknod`, `symlink`.

---

The `navigateFrom` also should be restructured to directly use the `ResourceAcquire` types. Trying to nest itself within `withTransactionF` becomes problematic because it is a recursive function that sometimes calles `this.navigate` and `this.navigateFrom`. It also mutates it's parameters as a sort of reference variables that is used across the recursions. A hold over from the original VFS design.

If we are locking we can only be locking the current directory, and once we navigate out, we need to release our lock.

The `_open` was non-recursive which was a bit easier, whereas here, we don't actually want to maintain transaction context under the recursive call graph. Therefore as soon as we do a "tail call here" we want to release the transaction for the current directory.

CMCDragonkai:
The `INodeManager` now exposes:

* `withTransactionF` & `withTransactionG` - convenience functions wrapping the the lockbox and transaction
* `withNewINodeTransactionF` & `withNewINodeTransactionG` - convenience functions augmenting `withTransactionF` & `withTransactionG` that allocates the inode using `inoAllocation()` resource, and locks it in the lockbox before locking more desired existing inodes and manages the transaction
* `inoAllocation` - the `ResourceAcquire` version of `INodeIndex`
* `transaction` - the `ResourceAcquire` version of `DBTransaction`

Functions in EncryptedFS that interact with these include:

* `createEncryptedFS` - uses only `withNewINodeTransactionF` at the beginning the root inode gets created
* `mkdir` - making directories involves using a while loop because of the `recursive` option
* `mknod` - uses only `withNewINodeTransactionF`
* `symlink` - uses only `withNewINodeTransactionF`
* `_open` - uses while loop due to following symlinks AND also due to `O_CREAT` option in which the target is created and still needs to be opened

All of the above methods except `createEncryptedFS` needs to consider concurrent operations racing to create the same inode. I've added this into `mkdir`, `mknod`, and `symlink`.

Also I've fixed a bunch of things in `mkdir`, there was many superfluous functionality in `mkdir` introduced in #56. Specifically added test for:

```ts
      // Asymmetric directory creation
      // the first promise will create dira and dira/dirb
      // the second promise will create dira/dirb/dirc
      await Promise.all([
        efs.mkdir('2/dira/dirb', { recursive: true }),
        efs.mkdir('2/dira/dirb/dirc', { recursive: true }),
      ]);
      expect(await efs.readdir('2/dira/dirb')).toStrictEqual(['dirc']);
      expect(await efs.readdir('2/dira/dirb/dirc')).toStrictEqual([]);
```

CMCDragonkai:
The `copyFile` method had a spurious `inoDeallocate`. This shouldn't be there. The destination being created is managed by `_open`, and it is not the job of `copyFile` to remove the inode. However Node FS does say it has no guarantees on the atomicity of `copyFile` and will try to remove the destination file if it was created for writing. We don't do this currently.

CMCDragonkai:
It looks like to if we are going to preserve the transaction context when the inode is already created as is now done in `mkdir` and `_open`, then the `navigateFrom` is going to need to be restructured to take a `tran` object.

Right now because the directory inode is created, and the transaction that is preserved already locks the containing directory and the new directory inode, when the loop iteration restarts and eventually tries to use `this.navigateFrom` on the new directory inode, this results in a deadlock because `navigateFrom` attempting to start a transaction on the same directory inode.

CMCDragonkai:
It does seem to make sense that one should be able to preserve the transactional context that `navigateFrom` will end up using. So it's worth checking that too.

CMCDragonkai:
With the new `tran` parameter to `navigateFrom`, I believe we have resolved the problem of `navigateFrom` starting multiple transaction contexts. It now all works.

Now it's time to tackle and finish off all the concurrency tests.

CMCDragonkai:
@tegefaulkes suggested that in the future, preventing race conditions of inode creation might be done by just locking the directory first. That might be simpler than our inode allocation resource management. Not sure atm.

Furthermore, if I move to getting pessimistic transactions implemented in `js-db`, this EFS may stick with how it is done now (dont fix what aint broke). But pessimistic transactions may make some of the transaction preservation mechanisms easier to understand.

CMCDragonkai:
For testing...

* concurrent inode creation
  - EncryptedFS.open
* concurrent file writes
  - EncryptedFS.write on multiple file descriptors
  - EncryptedFS.write on the same file descriptor
  - EncryptedFS.writeFile
  - EncryptedFS.appendFile

Some other ideas:

* test concurrent mkdir
* test concurrent symlink
* test concurrent mknod
* test concurrent open
* test mixed concurrent writes, mkdir, symlink, mknod
* test mixed reads and writes
* test mixed appends
* test mixed creation deletion
* Test start, create files, and stop, start again, and creating new inodes (that shouldn't conflict with existing data) (because of the iterator)
* tests with file descriptor

CMCDragonkai:
It looks like `rmdir` and `_rmdir` also is using separate transactions across their calls. They also have a `recursive` option similar to `mkdir`.

CMCDragonkai:
Currently `_rmdir` is only ever entered into if `recursive` option is true for `rmdir`. Nothing else calls it except itself.

It seems this was written incorrectly and should be done with a while loop as we have in `mkdir` as the loop is necessary for recursive directory creation.

CMCDragonkai:
In the node fs, the `rm` call is now available, and the `recursive` option is actually deprecated in the original `rmdir`. I can see how this makes sense, the `rmdir` doesn't make sense to be recursive, since removing anything else would be done differently. But `rm` would make sense to be capable of being recursive. The `rm` call isn't being used atm so we will leave that for later.

So I'm going to try for a while loop instead.

CMCDragonkai:
Because currently `navigation` via `EncryptedFS.navigate` is in its own transaction context, when it returns the inode index target as `navigated.target`, even if this is not null, when we then later acquire a transaction, locking on the `navigated.target`, after acquiring the lock, the target may already be deleted. This is because while waiting for the lock, it's possible that another call got the lock and deleted the inode.

So right now most of the methods deviates from the normal FS functionality because they will instead say `ENOTDIR` or `EISDIR` because they check if the type is correct, whereas if they were deleted concurrently should be throwing `ENOENT`.

Now this can be solved by just checking if the inode still exists always after we have acquired the lock. And if it doesn't, we throw ENOENT again.

But this code duplication is primarily due to the fact that `navigation` itself is in separate transaction context. In the future, it might make sense to change `navigate` to take a pessimistic transaction. So that once we create a transaction and pass it to navigate, it can acquire a lock on the navigated inode index. That way as long as you are navigating you have locked on the target. But that may change the `navigate` API.

So for now, I'll see how far I can go with just re-raising `ENOENT` if the inode no longer exists.

This also impacts the case where we are locking on the directory and target due to doing some directory manipulation. In this case, the directory itself may also be gone by the time we have navigated there. But if the directory itself is deleted, then any target inside the directory would also be deleted, so it should result in the same ENOENT error.

CMCDragonkai:
I've refactored the `rmdir` with `recursive` option. Further recursive testing is needed.

But I think alot of the methods that are just checking the target type after acquiring the transaction should also check if the target still exists, and throw an ENOENT after acquiring the lock, as this should help ensure a consistent set of errors during concurrent usage.

CMCDragonkai:
@tegefaulkes Will need you to help finish all the concurrency testing here.

1. All the methods which currently use transactions (most of them), right now need to have a check for whether the target exists after they have locked it. If the target doesn't exist, they should throw `ENOENT`. For example see `chroot`, it checks if the `target == null` and then also does it again after acquiring the transaction. This needs to be done almost all the other methods. You don't need to do this for `_open`, `mkdir`, `symlink`, `mknod`. Once this is done, it ensures that when there are concurrent operations where one of the operations ends up deleting the inode, all other calls waiting on locking that inode will end up throwing `ENOENT` instead of `ENOTDIR` or `EISDIR`. Please make sure to list all the methods you apply this change to, as I may some commentary for each of them. More details here: https://github.com/MatrixAI/js-encryptedfs/pull/63#issuecomment-1112931258
2. Additional `rmdir` tests should be written that test its `recursive` option under these situations:
   * When the parent directory lacks the write permission
   * When the parent directory lacks the search permission
   * When the target directory lacks write permission and is empty
   * When the target directory lacks write permission and is not empty
   * When the target directory lacks search permissions and is empty
   * When the target directory lacks search permissions and is not empty
   * When the target directory lacks write permissions, but a subdirectory under the target has write permissions and has files underneath them.
   The correct behaviour is to match however node's filesystem works. I suggest creating a prototype script that uses `fs/promises` and running the same test over the real filesystem, observing the side-effect and ensuring that our implementation does the same thing. In my recent changes to `rmdir` I believe this should be case, but new tests should be written.
3. More concurrent tests needs to be written. In the `tests/EncryptedFS.concurrent.test.ts`, I have started refactoring the concurrent tests. The uncommented tests should be currently working. All of the commented out tests needs to be incrementally uncommented and checked against EFS. Most importantly the uncommented tests are all too brittle. They should be testing the expected non-deterministic behaviour. That is if the observed side effects of 2 concurrent operations is the first then second, it is incorrect to test that the result is first then second. Instead the correct test is to allow for both first then second, and second then first. In some cases, one should use the `await sleep` used in a particular concurrent operation to force one operation to take longer than the other. Make sure to review my uncommented tests to understand how I want the tests to be structured. I find that using `Promise.allSettled` to quite expressive for these situations. Additional tests should be done (beyond just the commented out tests and incorporated into the describe structure):
   * concurrent mknod
   * concurrent symlink
   * concurrent mixed writes, mkdir, symlink, mknod
   * concurrent mixed readFile and writeFile
   * concurrent mixed read and write on multiple file descriptors
   * concurrent mixed read and write on the same file descriptor
   * concurrent mixed open and deletion
   * concurrent mixed stream read and write
   Correct behaviour should be based off node FS behaviour. Use a prototype script to test how node FS behaves.
4. INode allocation test, we need to test that the inode allocation is working since I updated this. Basically start the filesystem, write things to it, directories, files, take note of all the inodes allocated (delete some files and directories. Shutdown filesystem, recreate the filesystem (via EncryptedFS.start and EncryptedFS.createEncryptedFS) and ensure that the inode allocation reads from the FS. If the `iNodeCounter` is correct, then the next inode created should be the lowest deallocated inode index. Use the `getAll` to test this.

CMCDragonkai:
I've squashed most of the major concurrency fixes to the last WIP commit. So that commit will be our "concurrency fixing" commit.

tegefaulkes:
We're using `let variable` in a few places, This is causing us to loose type information and by extension proper type checking from typescript. I'll fix this up as I work.

tegefaulkes:
List of methods using transactions.
- chroot
- chdir
- access
- chmod
- chown
- copyFile
- fallocate
- fchmod
- fchown
- fstat
- ftruncate
- futimes
- lchmod
- lchown
- link
- lseek
- lstat
- read
- readdir
- readlink
- rename
- stat
- utimes
- symlink
- unlink
- mkdir
- _open
- rmdir
- navigateFrom

tegefaulkes:
What is the `syscall` parameter of `ErrorEncryptedFSError`?  We don't use it everytime we throw an `ErrorEncryptedFSError`. It seems to be a string of the method name that the error was thrown in. E.G 'access' when thrown in the `access` method. But in `chroot` it is 'readdir'. Is this correct?

CMCDragonkai:
I think that is purely descriptive, so you can change it as you see fit. We don't actually test for it in `expectError`. It's supposed to mean the "actual" syscall of the Linux kernel that gets called that had the error. So for EFS, it should just be the nearest EFS method call.

tegefaulkes:
Implementation of `read` seems wrong. It only holds the transaction for getting the files stat. But doesn't use the same transaction for actually reading the file. 

EncryptedFS.read

tegefaulkes:
In the case of `rename` we have a source and a target path. I'm not sure what to lock for the target path. The target inode will not exist so we can't lock that. Should I lock the parent of the target path? The Dir Inode in this case?

CMCDragonkai:
The `rename` might be tricky. You might need to lock all of these things:

1. The directory containing `source` and the directory containing `target`. This could be the same directory.
2. If the `target` is new inode index, it also need to be locked.
3. If the `target` is an existing inode, then you must lock that as well.

I have debated with myself on whether to lock the directory then target, or target than directory. I'm not sure what the implications are atm. So you'll see through testing.

tegefaulkes:
task reference
---
checklist:
- [ ] 1. All methods that use transactions need to check if target exists after it has locked. if the target doesn't exist after locking then the method needs to throw an `ENOENT` error. Use `chroot` as an example. Ignore this for `_open`, `mkdir`,`symlink` and `mknod`. List all methods affected, by extension make a checklist of all methods affected to tick off.
- [ ] 2. `rmdir` tests should be added for recursive option for the following conditions.
	- [ ] When the parent directory lacks the write permission
	- [ ] When the parent directory lacks the search permission
	- [ ] When the target directory lacks write permission and is empty
	- [ ] When the target directory lacks write permission and is not empty
	- [ ] When the target directory lacks search permissions and is empty
	- [ ] When the target directory lacks search permissions and is not empty
	- [ ] When the target directory lacks write permissions, but a subdirectory under the target has write permissions and has files underneath them. The correct behaviour is to match however node's filesystem works. I suggest creating a prototype script that uses `fs/promises` and running the same test over the real filesystem, observing the side-effect and ensuring that our implementation does the same thing. In my recent changes to `rmdir` I believe this should be case, but new tests should be written.
- [ ] 3. concurrency tests need to be updated. They need to be made non-deterministic since we can't determine the results, specifically order. Need to test expected outcome 1 OR outcome 2. since both are valid depending on order completed.
	- [ ] concurrent mknod
	- [ ] concurrent symlink
	- [ ] concurrent mixed writes, mkdir, symlink, mknod
	- [ ] concurrent mixed readFile and writeFile
	- [ ] concurrent mixed read and write on multiple file descriptors
	- [ ] concurrent mixed read and write on the same file descriptor
	- [ ] concurrent mixed open and deletion
	- [ ] concurrent mixed stream read and write
	- [ ] ftruncate test according to `./test-fail.ts` Correct behaviour should be based off node FS behaviour. Use a prototype script to test how node FS behaves.
- [ ] 4.  INode allocation test, we need to test that the inode allocation is working since I updated this. Basically start the filesystem, write things to it, directories, files, take note of all the inodes allocated (delete some files and directories. Shutdown filesystem, recreate the filesystem (via EncryptedFS.start and EncryptedFS.createEncryptedFS) and ensure that the inode allocation reads from the FS. If the `iNodeCounter` is correct, then the next inode created should be the lowest deallocated inode index. Use the `getAll` to test this.
- [ ] 5.  Remove all the `test-*` files (scaffolding).
- [ ] 6.  Ensure all comments are without `.` full stop, and don't capitalise the test description or message (this can be done incrementally)
- [ ] 7. Fix up any usage of `let variable` causing us to loose type information.


##### list of methods using transactions

- [X] chroot
- [X] chdir
- [X] access
- [X] chmod
- [X] chown
- [ ] copyFile - I think this is fixed?
- [X] fallocate
- [x] fchmod
- [X] fchown
- [X] fstat
- [X] ftruncate
- [X] futimes
- [X] lchmod
- [X] lchown
- [X] link - come back to this
- [X] lseek - come back
- [X] lstat
- [X] read
- [X] readdir
- [X] readlink
- [ ] rename - come back to this x2
- [X] stat
- [X] utimes
- [X] unlink
- [ ] rmdir - seems fine?
- [ ] navigateFrom - doesn't really need an ENOENT error?
- ~symlink~
- ~mkdir~
- ~ _open~
- ~mknod~

tegefaulkes:
There seems to be a bug when concurrently streaming writes to a file and truncating said file. Normally I'd expect that the size of a file when using `stat` should be equal or greater than the contents of the file. I've found a situation where after truncating and writing the file I end up with the stat size of 27 but the contents has a length of 45. 

This is likely a problem with the `EncryptedFS.createWriteStream` implementation. I'll look into it after finishing task 3.

tegefaulkes:
I've found another possible bug. When using `lseek` to move the cursor 20 bytes and then writing 4 bytes using `write` I'd expect the resulting file to have stat size 24 and the contents to have length 24 with the first 20 bytes filled with 0.

But I'm seeing a stat size of 24 and the contents of the file just the 4 bytes written by `write`. This is another thing I need to look into.

tegefaulkes:
General ETA for what is left.  refering to the [above](https://github.com/MatrixAI/js-encryptedfs/pull/63#issuecomment-1115512074) reference.
task 3. is hard to pin down. It's revealing small bugs as I progress so the overall time depends on things that are not fully known yet. But first blush i'm expecting up to 1 days to finish the tests and possibly 2 more to fix any bugs. For task 4. should be half a day if that at most. 5. and 6. are trivial and should take no time at all.
So overall
- 3. 3-ish days
- 4. 0.5 days
- 5. and 6. 1 hour.

Should be done within 4 days.

tegefaulkes:
Cleaned up the commits

CMCDragonkai:
The `expectReason` works against the promise settled result, while `expectError` works against promises. Would have preferred to call them something like `expectErrorPromise` and `expectErrorResult`, but we can proceed.

CMCDragonkai:
Note that `js-errors` is updated, but after this PR is merged, we want to update all the other packages again because they will be using the updated `js-errors` as well and we want all of our exceptions to be using the new `toJSON` and `fromJSON` serialisation/deserialisation.

PR #66, opened 2022-05-27
Upgrading DB to 4.0.2

emmacasolin:
### Description

Breaking changes in the DB affect how we use the iterator, but now allow us to use keys with zero bytes. The iterator now returns key paths instead of key buffers, so they need to be handled with this in mind.

### Tasks

- [x] 1. Fix all usages of db.iterator/tran.iterator

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@matrixai/async-init": "^1.7.3",
         "@matrixai/async-locks": "^2.2.4",
-        "@matrixai/db": "3.3.3",
+        "@matrixai/db": "^4.0.2",
         "@matrixai/errors": "^1.1.1",
         "@matrixai/logger": "^2.1.1",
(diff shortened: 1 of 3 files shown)


emmacasolin:
Many of the tests are failing after making changes. Changing the escape character in the DB to 0x00 fixes the majority of them, however, there are still issues with appending. Not sure yet what the cause of these are.

emmacasolin:
The db has been updated to 4.0.2 now, introducing key encoding to resolve the issues I was seeing here. The source of the test failures was the result of escaping the keys, which changed the lexicographic order that we rely on here in the EFS. Switching from escaping to encoding solves the problem and has been thoroughly tested in the DB so we can now continue here.

CMCDragonkai:
Great! Did you release the new version of EFS too?

PR #67, opened 2022-07-25
Updating js-db to 5.x.x (integrating RocksDB and DB-native concurrency control)

emmacasolin:
### Description
Updating CI/CD including:

- Continuous benchmarking
- Staging branch
- Advanced type checking

Related:

* https://github.com/MatrixAI/Polykey/pull/419
* https://github.com/MatrixAI/Polykey/issues/294
* https://github.com/MatrixAI/Polykey/issues/382

### Tasks
- [x] 1. Create new staging branch
- [x] 2. Synchronise changes from `js-logger`
- [X] 3. update `@matrixai/db` to  5.0.1,
- [X] 4. fix API changes for DB
- [x] 3. Fix failing tests
- [x] 4. Integrate fast-check and update any relevant tests.

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   "dependencies": {
     "@matrixai/async-init": "^1.8.2",
     "@matrixai/async-locks": "^3.1.2",
-    "@matrixai/db": "^5.0.1",
+    "@matrixai/db": "^5.0.2",
     "@matrixai/errors": "^1.1.3",
     "@matrixai/logger": "^3.0.0",
     "@matrixai/resources": "^1.1.4",
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "ts-node": "^10.9.1",
     "tsconfig-paths": "^3.9.0",
     "typedoc": "^0.22.15",
-    "typescript": "^4.5.2"
+    "typescript": "^4.5.2",
+    "fast-check": "^3.0.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   "dependencies": {
     "@matrixai/async-init": "^1.8.2",
     "@matrixai/async-locks": "^3.1.2",
-    "@matrixai/db": "^4.0.2",
+    "@matrixai/db": "^5.0.1",
     "@matrixai/errors": "^1.1.3",
     "@matrixai/logger": "^3.0.0",
     "@matrixai/resources": "^1.1.4",
--- a/src/inodes/INodeManager.ts
+++ b/src/inodes/INodeManager.ts
@@ -81,10 +81,9 @@ class INodeManager {
       await this.db.clear(this.mgrDbPath);
     }
     // Populate the inode counter with pre-existing inodes
-    for await (const [kP] of this.db.iterator(
-      { values: false },
-      this.iNodesDbPath,
-    )) {
+    for await (const [kP] of this.db.iterator(this.iNodesDbPath, {
+      values: false,
+    })) {
       this.iNodeCounter.allocate(inodesUtils.uniNodeId(kP[0] as INodeId));
     }
     // Clean up all dangling inodes that could not be removed due to references
@@ -122,10 +121,9 @@ class INodeManager {
    */
   protected async gcAll(): Promise<void> {
     await withF([this.db.transaction()], async ([tran]) => {
-      for await (const [kP] of tran.iterator(
-        { values: false },
-        this.gcDbPath,
-      )) {
+      for await (const [kP] of tran.iterator(this.gcDbPath, {
+        values: false,
+      })) {
         const ino = inodesUtils.uniNodeId(kP[0] as INodeId);
         // Snapshot doesn't need to be used because `this.gcAll` is only executed at `this.stop`
         const type = (await tran.get<INodeType>([
@@ -518,7 +516,7 @@ class INodeManager {
     tran: DBTransaction,
   ): Promise<void> {
     const dataPath = [...this.dataDbPath, ino.toString()];
-    for await (const [kP] of tran.iterator({ values: false }, dataPath)) {
+    for await (const [kP] of tran.iterator(dataPath, { values: false })) {
       await tran.del(dataPath.concat(kP));
     }
     await this.iNodeDestroy(ino, tran);
@@ -548,7 +546,7 @@ class INodeManager {
     } else {
       await this.dirUnsetRoot(tran);
     }
-    for await (const [kP] of tran.iterator({ values: false }, dirPath)) {
+    for await (const [kP] of tran.iterator(dirPath, { values: false })) {
       await tran.del(dirPath.concat(kP));
     }
     await this.iNodeDestroy(ino, tran);
@@ -640,14 +638,13 @@ class INodeManager {
       return yield* this.withTransactionG((tran) => this.getAll(tran));
     }
     // Consistent iteration on iNodesDbPath and gcDbPath
-    const gcIterator = tran.iterator<null>(
-      { valueAsBuffer: false },
-      this.gcDbPath,
-    );
+    const gcIterator = tran.iterator<null>(this.gcDbPath, {
+      valueAsBuffer: false,
+    });
     try {
       for await (const [inoData, type] of tran.iterator<INodeType>(
-        { valueAsBuffer: false },
         this.iNodesDbPath,
+        { valueAsBuffer: false },
       )) {
         const ino = inodesUtils.uniNodeId(inoData[0] as INodeId);
         gcIterator.seek(inoData);
@@ -659,7 +656,7 @@ class INodeManager {
         };
       }
     } finally {
-      await gcIterator.end();
+      await gcIterator.destroy();
     }
   }
 
@@ -931,8 +928,8 @@ class INodeManager {
       );
     }
     for await (const [kP, v] of tran.iterator<INodeIndex>(
-      { keyAsBuffer: false, valueAsBuffer: false },
       [...this.dirsDbPath, ino.toString()],
+      { keyAsBuffer: false, valueAsBuffer: false },
     )) {
       yield [kP.toString(), v];
     }
@@ -1100,10 +1097,10 @@ class INodeManager {
         }
       : { gte: inodesUtils.bufferId(startIdx) };
     let blockCount = startIdx;
-    for await (const [kP, v] of tran.iterator(options, [
-      ...this.dataDbPath,
-      ino.toString(),
-    ])) {
+    for await (const [kP, v] of tran.iterator(
+      [...this.dataDbPath, ino.toString()],
+      options,
+    )) {
       // This is to account for the case where a some blocks are missing in a database
       // i.e. blocks 0 -> 3 have data and a write operation was performed on blocks 7 -> 8
       while (blockCount < inodesUtils.unbufferId(kP[0] as BufferId)) {
@@ -1131,10 +1128,10 @@ class INodeManager {
     }
     const options = { limit: 1, reverse: true };
     let key, value;
-    for await (const [kP, v] of tran.iterator(options, [
-      ...this.dataDbPath,
-      ino.toString(),
-    ])) {
+    for await (const [kP, v] of tran.iterator(
+      [...this.dataDbPath, ino.toString()],
+      options,
+    )) {
       key = inodesUtils.unbufferId(kP[0] as BufferId);
       value = v;
     }
--- a/src/inodes/INodeManager.ts
+++ b/src/inodes/INodeManager.ts
@@ -14,7 +14,6 @@ import {
   CreateDestroyStartStop,
   ready,
 } from '@matrixai/async-init/dist/CreateDestroyStartStop';
-import { Lock, LockBox } from '@matrixai/async-locks';
 import { withF, withG } from '@matrixai/resources';
 import Counter from 'resource-counter';
 import * as inodesUtils from './utils';
@@ -68,7 +67,6 @@ class INodeManager {
   protected iNodeCounter: Counter = new Counter(1);
   protected iNodeAllocations: Map<string, Ref<INodeIndex>> = new Map();
   protected refs: Map<INodeIndex, number> = new Map();
-  protected locks: LockBox<Lock> = new LockBox();
 
   constructor({ db, logger }: { db: DB; logger: Logger }) {
     this.logger = logger;
@@ -208,25 +206,9 @@ class INodeManager {
     ...inos: Array<INodeIndex>
   ): ResourceAcquire<DBTransaction> {
     return async () => {
-      const locksAcquire = this.locks.lock(
-        ...inos.map<[INodeIndex, typeof Lock]>((ino) => [ino, Lock]),
-      );
-      const transactionAcquire = this.db.transaction();
-      const [locksRelease] = await locksAcquire();
-      let transactionRelease, tran;
-      try {
-        [transactionRelease, tran] = await transactionAcquire();
-      } catch (e) {
-        await locksRelease();
-        throw e;
-      }
-      return [
-        async () => {
-          await transactionRelease();
-          await locksRelease();
-        },
-        tran,
-      ];
+      const [transactionRelease, tran] = await this.db.transaction()();
+      await tran!.lock(...inos);
+      return [transactionRelease, tran];
     };
   }
 
@@ -238,13 +220,10 @@ class INodeManager {
     ]
   ): Promise<T> {
     const f = params.pop() as (tran: DBTransaction) => Promise<T>;
-    const lockRequests = (params as Array<INodeIndex>).map<
-      [INodeIndex, typeof Lock]
-    >((ino) => [ino, Lock]);
-    return withF(
-      [this.db.transaction(), this.locks.lock(...lockRequests)],
-      ([tran]) => f(tran),
-    );
+    return await this.db.withTransactionF(async (tran) => {
+      await tran.lock(...(params as Array<INodeIndex>));
+      return f(tran);
+    });
   }
 
   @ready(new inodesErrors.ErrorINodeManagerNotRunning())
@@ -257,13 +236,10 @@ class INodeManager {
     const g = params.pop() as (
       tran: DBTransaction,
     ) => AsyncGenerator<T, TReturn, TNext>;
-    const lockRequests = (params as Array<INodeIndex>).map<
-      [INodeIndex, typeof Lock]
-    >((ino) => [ino, Lock]);
-    return withG(
-      [this.db.transaction(), this.locks.lock(...lockRequests)],
-      ([tran]) => g(tran),
-    );
+    return this.db.withTransactionG(async function* (tran) {
+      await tran.lock(...(params as Array<INodeIndex>));
+      return yield* g(tran);
+    });
   }
 
   @ready(new inodesErrors.ErrorINodeManagerNotRunning())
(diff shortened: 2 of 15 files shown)


emmacasolin:
The problem with updating js-logger here (which we need to do in order to update it in PK) is that it also means we need to update the DB, but this also means incorporating all of the new (breaking) changes from 5.0.0 of `js-db`. I think I've made the correct changes to the source code, but there are several failing tests. This is to be expected though, since `js-db` has undergone some major changes. Most of the failures are to do with concurrency, which makes sense given the focus of most of the recent changes to the db.

CMCDragonkai:
Hold on this PR, the js-db update is more pressing first. And we don't need to update the EFS yet. The js-db 5.0.0 is first being updated in PK. EFS can stay on the original js-db.

CMCDragonkai:
What you can do is to merge in the non-js-db and non-js-logger changes in first. There's just a bunch of CI/CD stuff.

Can you split up the commit into a separate PR for just the staging and ci/cd and other chores.

emmacasolin:
> Hold on this PR, the js-db update is more pressing first. And we don't need to update the EFS yet. The js-db 5.0.0 is first being updated in PK. EFS can stay on the original js-db.

Ok. EFS won't work inside PK if it's running the old DB though, since the old DB uses the old logger, which conflicts with the new logger used by the new DB in PK, so it will need to be updated here eventually.

CMCDragonkai:
Actually you can do it. You just won't be able to pass a "child" logger into the DB that EFS is using. So you can just remove that parameter entirely.

emmacasolin:
Now that I've updated the logger for all of the other libraries though we can no longer pass a logger through to any of the other libraries from EFS. I think this is only an issue for `js-workers` though, which is used only in the benchmarks and in one of the tests (`efsWorker.test.ts`), so we could just not pass a logger into those.

emmacasolin:
https://github.com/MatrixAI/js-encryptedfs/pull/67/commits/ee0b9e2171a9cf89f4905cbff7058e1ba8924f39 now contains all of the CI/CD changes but doesn't update the logger or DB. As such, it doesn't pass a logger through to any of the workers since `js-workers` uses a different version of `js-logger`. All of the tests should be passing on that commit, but they don't pass on the most recent commit due to concurrency changes from the latest DB.

CMCDragonkai:
Once #68 is merged, rename this PR to focus on js-db 5.x upgrade.

CMCDragonkai:
@emmacasolin please update to using 1.3.6 of js-workers. It is now compatible with swc.

The quickfix is to swap from using `new X();` to `new this();` wherever we have static async creators.

This needs to be fixed in `EncryptedFS.ts`. Change to:

```
    const efs = new this({
```

To make it usable. Then `swc` should be enabled.

CMCDragonkai:
The same fix has to be applied to PK as well under `staging` branch. Can you do it at the same time as you do you js-logger updates there.

tegefaulkes:
008457 fixes #69. There still are failing tests in this PR but it passed when tested on top of staging.

tests/fd/FileDescriptor.test.ts
```diff
@@ -22,7 +21,7 @@ describe('File Descriptor', () => {
   const origBuffer = Buffer.from('Test Buffer for File Descriptor');
   beforeEach(async () => {
     dataDir = await fs.promises.mkdtemp(
-      path.join(os.tmpdir(), 'encryptedfs-test-'),
```
CMCDragonkai, line 25: Actually these are meant to still use `path.join` and not `pathJoin`. This is because, this is a real directory, and not inside EFS.

CMCDragonkai:
Creation of temporary directory with `os.tmpdir()` should still be using `path.join` as normal.

tests/fd/FileDescriptor.test.ts
tegefaulkes: Shoot, my bad. I'll fix that real quick.

CMCDragonkai:
@emmacasolin this PR needs to be split up. The js-db changes should be moved into a new PR.

This PR should only focus on synchronising changes, updating the js-logger, and the fix that @tegefaulkes just added.

CMCDragonkai:
I suspect that the test failures are coming from the js-db changes. The js-db update should be postponed till later. @tegefaulkes

CMCDragonkai:
Actually I was confused, the recent fix should go to #68 and #67. Will rename this PR to suit.

@tegefaulkes can you reset the commit you just pushed?

tests/fd/FileDescriptor.test.ts
CMCDragonkai: Should be done on #68.

CMCDragonkai:
When this is being worked on all:

```
      // @ts-ignore - version of js-logger is incompatible (remove when js-db updates to 5.* here)
```

Should be removed from the source as `js-logger` will be compatible and we'd have completed the upgrade.

Unlike in Polykey where it has a commit that we can revert, this PR won't have that convenience.

tegefaulkes:
Removed the fix commit from here and pushed it to staging.

tegefaulkes:
Fast-check can be used to combine a number of concurrency tests together. For example, concurrent file writes using a number of methods can be the same test. for these tests the the result matters less than avoiding transaction conflicts. we still want to check that something has been written however.

tegefaulkes:
The `INodeManager` is using a `LockBox` along side the the database transactions. I think  we will want to make use of the transaction locks now.

`FileDescriptor` has its own `Lock` for locking on cursor updates. This one should be left alone.

tegefaulkes:
looking at the `EncyptedFs > write > can copy files` test failure. it has to create the destination file and the destination doesn't exist after creation causing the issue here. Seems that copyFile uses `this._open` to open a FD AND create the file. but this is not composed with the wider transaction thus causing the problem.

I'm sure that `_open` is used inside of other methods such as `appendFile` and `open`. I think the problem only happens when `_open` has to create the file when opening it. In any case we need to be able to compose it with larger transactions.

CMCDragonkai:
1. Yes swap to use transaction locks instead of `LockBox`.
2. Where `INodeManager` is using `tran` as the first parameter, swap it to the last parameter in preparation for the decorator usage
3. When an fd is opened, it has to be opened to a created inode.

tegefaulkes:
The `INodeManager` is now using the transaction locks. So far as I can tell, the `tran` parameter is already the last parameter where it's used.

I'm not sure what you mean by point 3.

tegefaulkes:
This may be useful to streamline fast-check with jest.
https://www.npmjs.com/package/@fast-check/jest

CMCDragonkai:
For point 3, I just mean when fds are opened. They are always opened to a existing inode. If a fd is alive, inode is alive. You can delete names, but you cannot delete inodes until all links are gone, that includes fds and other names (aka hardlinks).

So yes when opening a fd, it has to check if it needs to create an inode at the same time. So race conditions can occur here due to concurrent creation.

tegefaulkes:
The problem here was that `copyFile` uses `_open` to create a fd pointing to the destination. It's using the create flag so that the file is created if it didn't exist. At the time this was done in a separate transaction so the `copyFile`'s overall transaction didn't see the file get created resulting in an error.

tegefaulkes:
For the concurrency tests, we should only really check for if 1 method succeeds and others fail. We shouldn't really care what the reason is besides that it was specifically a `ErrorEncryptedFSError` and not a transaction conflict. This way I can use the fast-check scheduler to simplify and combine tests while ending up with more robust tests.

tegefaulkes:
Is it really worth checking the result of a concurrent write? The expectation is that the writes should complete without error. Should not throw a transaction conflict. What is actually written to the file is technically undefined behaviour so checking what is written feels like unnecessary complexity added to the test. We only really need to check that the number of bytes read is within the expected range.

tegefaulkes:
re-based on staging.

CMCDragonkai:
> Is it really worth checking the result of a concurrent write? The expectation is that the writes should complete without error. Should not throw a transaction conflict. What is actually written to the file is technically undefined behaviour so checking what is written feels like unnecessary complexity added to the test. We only really need to check that the number of bytes read is within the expected range.

Yes in terms of checking, you can apply a predicate check to the patterns of the resulting concurrent write. This way during fast check, on every iteration they do, which they are changing up the concurrent operation order, we get a good concurrent coverage. If our predicates fail for a certain order, these tests will tell us.

The predicate can be a regular expression. That's how I wrote it already, but I lacked some sort of async scheduler to actually exhaustively check concurrent coverage. Fast check should be capable of exhausting the variations.

CMCDragonkai:
Also we don't need to change over all the tests. Just fix any of the concurrent tests and add any additional ones to ensure we have consistent behaviour especially relating to transaction conflicts.

The EFS is just getting more bug free, but it will be more important to work together on integrating the tracing and queueing to address testnet tests starting tomorrow.

tegefaulkes:
This should be good to merge now. Pending review if desired.

CMCDragonkai:
Can you rerun the benchmarks and show the comparison here. Since the db has changed, I'm expecting it should be alot faster.

src/inodes/INodeManager.ts
```diff
@@ -68,7 +67,7 @@ class INodeManager {
   protected iNodeCounter: Counter = new Counter(1);
   protected iNodeAllocations: Map<string, Ref<INodeIndex>> = new Map();
   protected refs: Map<INodeIndex, number> = new Map();
-  protected locks: LockBox<Lock> = new LockBox();
+  // Protected locks: LockBox<Lock> = new LockBox();
```
CMCDragonkai, line 70: This comment can be removed.

package.json
```diff
@@ -46,7 +46,7 @@
   "dependencies": {
     "@matrixai/async-init": "^1.8.2",
     "@matrixai/async-locks": "^3.1.2",
```
CMCDragonkai, line 48: Is this dependency directly needed? If not, we can remove.

tegefaulkes: Are you reffering to locks or both init and locks? The `Lock` is still used by FD operations for locking cursor updates.

src/inodes/INodeManager.ts
```diff
@@ -210,25 +207,11 @@ class INodeManager {
...
-        throw e;
-      }
-      return [
-        async () => {
-          await transactionRelease();
-          await locksRelease();
-        },
-        tran,
-      ];
+      return [transactionRelease, tran];
```
CMCDragonkai, line 214: Do we need this function anymore...?

Can this be changed to using `withF`... rather than `inodeMgr.transaction`.

Just wondering if we should preserve this old style of API.

src/inodes/INodeManager.ts
```diff
@@ -289,17 +273,15 @@ class INodeManager {
     if (typeof params[0] !== 'number') {
       navigated = params.shift() as Readonly<{ dir: INodeIndex; name: string }>;
     }
-    const lockRequests = (params as Array<INodeIndex>).map<
-      [INodeIndex, typeof Lock]
-    >((ino) => [ino, Lock]);
+    const lockRequests = (params as Array<INodeIndex>).map((iNode) =>
+      iNode.toString(),
```
CMCDragonkai, line 277: Is this necessary? I mean I forgot whether the existing locking system in DB already converts to string naturally.

package.json
CMCDragonkai: Just `async-locks`. I see ok.

src/inodes/INodeManager.ts
tegefaulkes: I can inline it, but it's a nice wrapper for locking the inodes along side the transaction.

src/inodes/INodeManager.ts
tegefaulkes: It should, but i encountered a problem with it. I'll check it again.

tegefaulkes: Trying to do a `tran.lock(...inos)` gives us the type error `TS2345: Argument of type 'INodeIndex' is not assignable to parameter of type 'string | [key: ToString, type?: "read" | "write" | undefined, timeout?: number | undefined]'.   Type 'INodeIndex' is not assignable to type '[key: ToString, type?: "read" | "write" | undefined, timeout?: number | undefined]'.`

Using `await tran!.lock([inos[0], 'write'])` works but we'd have to map it to the `[key, 'write']` form anyway. seems like just using a key won't auto covert to a string. Seems like the API needs to be updated tp `ToString | string | ect`?

CMCDragonkai:
Oh yea you need to provide the type of the lock. So don't worry.

On 19 August 2022 11:49:15 am AEST, Brian Botha ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>
>
>
>
>> @@ -289,17 +273,15 @@ class INodeManager {
>
>     if (typeof params[0] !== 'number') {
>
>       navigated = params.shift() as Readonly<{ dir: INodeIndex; name: string }>;
>
>     }
>
>-    const lockRequests = (params as Array<INodeIndex>).map<
>
>-      [INodeIndex, typeof Lock]
>
>-    >((ino) => [ino, Lock]);
>
>+    const lockRequests = (params as Array<INodeIndex>).map((iNode) =>
>
>+      iNode.toString(),
>
>
>
>Trying to do a `tran.lock(...inos)` gives us the type error `TS2345: Argument of type 'INodeIndex' is not assignable to parameter of type 'string | [key: ToString, type?: "read" | "write" | undefined, timeout?: number | undefined]'.   Type 'INodeIndex' is not assignable to type '[key: ToString, type?: "read" | "write" | undefined, timeout?: number | undefined]'.`
>
>
>
>Using `await tran!.lock([inos[0], 'write'])` works but we'd have to map it to the `[key, 'write']` form anyway. seems like just using a key won't auto covert to a string. Seems like the API needs to be updated tp `ToString | string | ect`?
>
>
>
>-- >
>Reply to this email directly or view it on GitHub:
>
>https://github.com/MatrixAI/js-encryptedfs/pull/67#discussion_r949729815
>
>You are receiving this because you commented.
>
>
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/inodes/INodeManager.ts
tegefaulkes: This is used in 5 different places in a weird way. it's not used with a `WithF` block in any of these places so a closer look is needed before removing this. Might not be worth the time right now.

tests/EncryptedFS.concurrent.test.ts
```diff
@@ -23,6 +24,23 @@ describe(`${EncryptedFS.name} Concurrency`, () => {
...
   let iNodeMgr: INodeManager;
   let efs: EncryptedFS;
+
+  const gateFactory = (s: fc.Scheduler) => {
+    const scheduledStarter = s.scheduleFunction(async () => {});
+    return async <T>(f: () => Promise<T>) => {
+      await scheduledStarter();
+      return await f();
+    };
+  };
```
CMCDragonkai, line 34: We should find a better way to apply barriers to the scheduling system of fast-check. Should search through the issues.

tests/EncryptedFS.concurrent.test.ts
```diff
@@ -57,6 +74,42 @@ describe(`${EncryptedFS.name} Concurrency`, () => {
...
+              true,
+            );
+          })
+          .afterEach(async () => {
+            // Cleaning up
+            await efs.rmdir('dir', { recursive: true });
+          }),
+        { numRuns: 50 },
+      );
+    });
```
CMCDragonkai, line 112: Should always apply the `globalThis.defaultTimeout` here. The jest library version of fast-check doesn't have this by default. On the CI/CD machines they may run this very slowly.

We should apply a "max timeout" to avoid running past the time limit on the CI/CD machines.

You could do something lower than `defaultTimeout`... like `globalThis.defaultTimeout - 1000`.

tests/EncryptedFS.concurrent.test.ts
```diff
@@ -240,120 +293,135 @@ describe(`${EncryptedFS.name} Concurrency`, () => {
...
+          ).toHaveLength(3);
+          for (const fd of fds) {
+            await efs.close(fd);
+          }
+        }),
+        { numRuns: 10 },
       );
-      for (const fd of fds) {
-        await efs.close(fd);
-      }
```
CMCDragonkai, line 276: This is replacement test does not cover the same semantics. It's missing the:

> Concurrent writes of different length results in "last write wins" or a merge

You should have both here. Also retain the comments in the resulting test rewrite.

tests/EncryptedFS.concurrent.test.ts
```diff
@@ -240,120 +293,135 @@ describe(`${EncryptedFS.name} Concurrency`, () => {
...
+          const prom = Promise.all([
+            gate(() => efs.write(fds[0], contents[0])),
+            gate(() => efs.write(fds[1], contents[1])),
+          ]);
+          await s.waitAll();
+          await prom;
+
+          expect(
+            await efs.readFile('test', { encoding: 'utf-8' }),
+          ).toHaveLength(3);
```
CMCDragonkai, line 314: Change this to a regular expression. Like` expect().toMatch(/one|two/)`.

CMCDragonkai: Use regular expressions for any kind of non-deterministic results.

tests/EncryptedFS.concurrent.test.ts
CMCDragonkai: We don't actually want a fixed number of runs. It should be based on time.

CMCDragonkai: The fixed number of runs can be a "maximum" variants we expect. But the timeout should be hard limit.

tests/EncryptedFS.concurrent.test.ts
```diff
@@ -240,120 +293,135 @@ describe(`${EncryptedFS.name} Concurrency`, () => {
...
+            gate(() => efs.write(fd, 'aaa')),
+            gate(() => efs.write(fd, 'bbb')),
+          ]);
+          await s.waitAll();
+          await prom;
+
+          expect(
+            await efs.readFile('test', { encoding: 'utf-8' }),
+          ).toHaveLength(6);
+          await efs.close(fd);
```
CMCDragonkai, line 339: This should be a regular expression.

```
expect().toMatch(/aaabbb|bbbaaa/)
```

CMCDragonkai:
Ok so for all the concurrent tests where you have replaced the test, change to using a regular expression to match the original predicates on the output. I've given some examples above.

For the new tests, this can also be done if you see an easy way to write the regular expression. Prefer this over simply checks like `3`.

Then proceed to merge to `staging`.

On the staging branch, regenerate docs and redo the benchmarks.

tests/EncryptedFS.concurrent.test.ts
CMCDragonkai: I think there is way to do this properly, see: https://github.com/dubzzz/fast-check/issues/564.

And the associated commits.

CMCDragonkai:
There's a documented way of doing delaying the actual async functions for the fast check scheduler: https://github.com/dubzzz/fast-check/issues/564. See the relevant commits. I also asked a question upstream to see how it should be done.

src/inodes/INodeManager.ts
tegefaulkes: WE determined that this is fine as it is.

tests/EncryptedFS.concurrent.test.ts
tegefaulkes: Replaced the `gateFactory` with the `scheduleCall` implementation. This adds the ability to label the calls for clearer test failures.

CMCDragonkai: I suggest moving this into `tests/utils`.

tests/EncryptedFS.concurrent.test.ts
tegefaulkes: I've added timeouts and adjusted the `numRuns` for the `fast-check` tests.

tests/EncryptedFS.concurrent.test.ts
tegefaulkes: I've added the check back in.

tests/EncryptedFS.concurrent.test.ts
tegefaulkes: I've added the check back in.

tests/EncryptedFS.concurrent.test.ts
tegefaulkes: I've added the check back in.

tegefaulkes:
RIght, just need to squash and checks. Then this will be ready to merge.

CMCDragonkai:
Did you check the rest of the tests you replaced? I didn't go through all of them.

tegefaulkes:
Yes, I went through all of the ones I updated.

tegefaulkes:
Cleaned up the commit history.

tegefaulkes:
Just waiting on CI now.

CMCDragonkai:
I've asked about the integrating timeouts between fast check and jest: https://github.com/dubzzz/fast-check/issues/3084

PR #81, opened 2023-06-26
ci: merge staging to master

maxwell-aiden:
This is an automatic PR generated by the pipeline CI/CD. This will be automatically fast-forward merged if successful.


CMCDragonkai:
The windows failures revealed the problem is that we were not stopping the DB in concurrent tests because we no longer manage the lifecycle of managed DBs.

I also realised that this also applies to `iNodeMgr`, so I changed to using `dbOwned` and `iNodeMgrOwned`.

This may impact downstream tests and usage of EFS @tegefaulkes.

Beware of this @amydevs

CMCDragonkai:
Pushing 3.5.8, 3.5.7 will be lost.

